PR #174, opened 2024-02-01
Attempt implementing background CLI flag for Install/Uninstall

### LINKED ISSUE #133: Add CLI-Flag to not await install/uninstall commands

- [x] install command
- [ ] uninstall command

jalavosus:
Hey there! I saw #133 and figured, "I could do this in my spare time." So, I started on it. 

This PR is currently a draft, as I haven't added the bool flag to anything other than cmd/install. It's also a draft as I'm not the best at writing "user-friendly" flag usage strings, so input on that in general would be lovely. 

Overall, the PR modifies (or, will modify) Install and Uninstall to take an optional boolean flag "-b/--background" (name can obviously be changed), which, if passed and `true`, will run Install/Uninstall and not wait for a status to be returned. 

In either this PR or a separate one, it might make sense to add a `Status` subcommand to Install and Uninstall to check on background un/install status - let me know if that's a thing I should work on here, or separately - and of course, at all.

Thanks! 

Closes #133 

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -5,13 +5,22 @@ import (
 	"os"
 	"strings"
 
+	"github.com/spf13/cobra"
+
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/repo"
 	"github.com/glasskube/glasskube/pkg/client"
 	"github.com/glasskube/glasskube/pkg/condition"
 	"github.com/glasskube/glasskube/pkg/install"
-	"github.com/spf13/cobra"
 )
+
+type installOptions struct {
+	background bool
+}
+
+var installCmdOptions = installOptions{
+	background: false,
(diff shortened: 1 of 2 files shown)


christophenne (commented):
Hi @jalavosus,

thanks for your contribution, really appreciate it!

* For me b/background sounds fine.
* Regarding the status subcommand, I'm not sure yet if we need it - there is the `glasskube ls` command already, which delivers the same information (Ready/Pending/Failed) - except of course for all packages. If we want to pin it down to one specific package, I would rather have this as a parameter of the `ls` command instead. `glasskube install status cert-manager` seems strange as it doesn't really install it. Of course open for discussion.

pmig:
Hello @jalavosus, I hope you're doing well. I wanted to check in regarding your pull request. Are you still actively working on it, or would you like me to take over? Your contributions are greatly appreciated either way. Let me know what works best for you. Thank you!

PR #204, opened 2024-02-05
Fix : Removed logger dependency issue from the repo client(Issue #99)

### LINKED ISSUE #99: Remove logger dependency from repo client
There is a logging-workaround, which should be refactored, see https://github.com/glasskube/glasskube/pull/97#discussion_r1469395170

yrs147:
Closes #99  

## 📑 Description
Removed the logger dependency from repo client

## ✅ Checks


yrs147:
@christophenne

christophenne:
Hi @yrs147 , thanks for your PR.
Please have a look at the failing checks, one is that the `ctx` variable is now unused - you can remove it. 
Also the commit message is not allowed to have upper case characters (Fix -> fix, Remove -> remove).

Additionally, please remove setting the null logger in `pkg/client/client.go` in `SetupContext`, since this is what we wanted to get rid of. 

If there are any questions, feel free to ask.

yrs147:
Thanks for guiding @christophenne , I've made the necessary changes , can you please take a look at it

PR #208, opened 2024-02-06
feat: add `glasskube open` cmd

### LINKED ISSUE #107: Support port-forwarding and proxying to kubernetes applications
The client package should provide functionality to make a deployed application available via an URL (e.g. proxy to the kubernetes-dashboard, portforwarding, ...).

### LINKED ISSUE #108: `glasskube open` command
The CLI should support the user in opening a kubernetes application in the browser. 

Example:

`glasskube open cyclops` should take care of the port-forwarding to get a URL to the application, which is then automatically opened in the browser and printed into the console as a fallback.

Related: #107

kosmoz:
Closes #107
Closes #108

## 📑 Description
This PR adds the `glasskube open` subcommand.

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/internal/cliutils/browser.go
+++ b/internal/cliutils/browser.go
@@ -1,0 +1,22 @@
+package cliutils
+
+import (
+	"fmt"
+	"os/exec"
+	"runtime"
+)
+
+func OpenInBrowser(url string) error {
+	var err error
+	switch runtime.GOOS {
+	case "linux":
+		err = exec.Command("xdg-open", url).Start()
+	case "windows":
+		err = exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
+	case "darwin":
+		err = exec.Command("open", url).Start()
+	default:
+		err = fmt.Errorf("unsupported platform")
+	}
+	return err
+}
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -1,0 +1,12 @@
+package open
+
+import (
+	"context"
+
+	"github.com/glasskube/glasskube/pkg/client"
+)
+
+func Portforward(client *client.PackageV1Alpha1Client, ctx context.Context, packageName string) error {
+	// portforward.New(httpstream.Dialer())
+	return nil
+}
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -3,6 +3,7 @@ package open
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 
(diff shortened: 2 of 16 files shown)


pkg/open/open.go
```diff
@@ -0,0 +1,238 @@
...
+	result := openResult{opener: o}
+	var futures []future.Future
+	for _, entrypoint := range entrypoints {
+		if entrypointName == "" || entrypoint.Name == entrypointName {
+			e := entrypoint
+			readyCh := make(chan struct{})
+			stopCh := make(chan struct{})
+			o.readyCh = append(o.readyCh, readyCh)
+			o.stopCh = append(o.stopCh, stopCh)
+			entrypointFuture, err := o.open(ctx, packageInfo.Status.Manifest, e, readyCh, stopCh)
```
pmig, line 78: A port conflict will lead to a silent failure. (e.g. executing `docusaurus start` will listen on port 3000 and `glasskube open cyclops` will also try to create a port forward for this specific port). As of now this is a silent error and the docusaurus website is just opened in the browser. We should exit the open command with an error.

pkg/open/open.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+		if !exists {
+			return nil, fmt.Errorf("package %v has no entrypoint %v", packageName, entrypointName)
+		}
+	}
+
+	result := openResult{opener: o}
+	var futures []future.Future
+	for _, entrypoint := range entrypoints {
+		if entrypointName == "" || entrypoint.Name == entrypointName {
```
pmig, line 72: cyclops requires the backend _and_ the frontend service to be get port-forwarded in other to work correctly. The backend actually get called via [`localhost:8080`](https://github.com/cyclops-ui/cyclops/blob/5637e47724be140481150767683239ef8abdd031/install/cyclops-install.yaml#L257)

We might need to forward multiple entrypoints, but only the first entry-point will get opened?  (e.g. `glasskube open cyclops ui ctrl`)

pmig (changes requested):
It worked! (after stopping `docusaurus`)

That's a really nice feature that looks well written! I made 3 comments inline.

internal/cliutils/browser.go
```diff
@@ -0,0 +1,22 @@
+package cliutils
+
+import (
+	"fmt"
+	"os/exec"
+	"runtime"
+)
+
+func OpenInBrowser(url string) error {
```
christophenne, line 9: please also make the server use this function and delete the unnecessary implementation there

pkg/open/open.go
kosmoz: Yes, this is true. That's why, by default, all entrypoints are forwarded and the first one is opened in the browser. This works for cyclops, once https://github.com/glasskube/packages/pull/13 gets merged.

I implemented it this way so we don't have to mess with the schema too much …

pkg/open/open.go
kosmoz: That's interesting, because it does that for me…
image

kosmoz: I introduced an additional check for `:[port]` to catch this problem.

pmig (approved):

christophenne (approved):

Issue #172, opened 2024-02-01
Show Package Details on separate page


erikauer:
The package overview is great, but I'd like to see a more detailed description to the packages and additional information like links to the website or github. I would expect a separate detail page for each package that can be entered from the overview.

aatbip:
I want to take on this issue.

pmig:
@aatbip  sure. If there is a need to extend the package manifest we are happy to go over your suggestions.

pmig:
@aatbip sadly this issue is also related to #126 which we need to tackle first in order to proceed with this one, please feel free to select another issue from this list: https://github.com/glasskube/glasskube/labels/help%20wanted

PR #216, opened 2024-02-07
feat(ui): add real-time updates #164 and refactor #126

### LINKED ISSUE #126: UI Revamp - Introduce htmx.org
For the MVP we took some shortcuts to get the UI presentable quickly (e.g. no real build steps, full page reloads, duplicating templates and re-usable components, copying the CSS theme and inline style definitions). 

Before adding too many new features on top of this prototype, we should think about the UI in the long-term from a technology and architecture perspective, taking into account the upcoming features.  
* Do we want to continue with server-side rendered templates? If so, how do we manage more interactive use-cases?
* Do we need application state on the client?

This is to be decided and then implemented.

### LINKED ISSUE #164: Refresh installation status on the UI automatically
The package overview page should update itself when a package installation is finished, instead of having the user manually refresh it.

christophenne:
Closes #164
Closes #126

## 📑 Description
This PR
* refactors the existing UI solution to use sub-templates/components to make the markup more reusable
* introduces htmx to tackle dynamic partial page updates
* introduces websockets to push down updates to the client

## ✅ Checks

## ℹ Additional Information
Websocket code is based on the gorilla examples, e.g. https://github.com/gorilla/websocket/blob/main/examples/chat/hub.go


internal/web/server.go
```diff
@@ -38,60 +47,102 @@ type ServerConfigSupport struct {
...
+	loadTemplates()
+}
+
+func loadTemplates() {
+	templateFuncs := template.FuncMap{
+		"ToInstallButtonInput": components.ToInstallButtonInput,
 	}
+	baseTemplate = template.Must(
+		template.New("base.html").Funcs(templateFuncs).ParseFS(embededFs, "templates/layout/base.html"))
+	pkgsPageTmpl = template.Must(template.Must(baseTemplate.Clone()).
```
christophenne, line 71: I'm pretty sure this (templates location, parsing, nesting, splitting pages/components) needs more polishing, but I would continue it in #172, because there we have the case of adding another page + probably re-using the button again. And I think at this point it makes more sense to "craft" some more generic system than it is now.

kosmoz: This might be a stupid question but why do we need different template instances for every page/component?

christophenne: To be honest I couldn't make it work any other way, but I can look into it again. What would you have expected instead? One `tmpl` instance that is used for everything?

kosmoz: Yes, something like 
```
tpl := template.ParseFS(fs, "templates/pages/*.html", "templates/components/*.html", "...")
```

kosmoz: because you still have to specify the "entrypoint" anyways…

internal/web/server.go
```diff
@@ -38,60 +47,102 @@ type ServerConfigSupport struct {
...
+		fmt.Fprintf(os.Stderr, "Failed to create WebSocket client: %v", err)
+		return
 	}
-	supportTemplate, err := template.ParseFS(embededFs, "templates/support.html")
-	if err != nil {
-		return err
+	h.Register <- wsClient
+	go wsClient.HandleReads()
+	go wsClient.HandleWrites()
+}
```
kosmoz, line 59: I would move this to hub.go and rename to `handle` or something similar.

internal/web/server.go
```diff
@@ -38,60 +47,102 @@ type ServerConfigSupport struct {
...
+				InstallBlocking(ctx, pkgName)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "An error occurred rendering the response: \n%v\n", err)
+				fmt.Fprintf(os.Stderr, "An error occurred installing %v: \n%v\n", pkgName, err)
 			}
+
+			// broadcast the status update to all clients
+			var bf bytes.Buffer
+			components.RenderInstallButton(&bf, installBtnTmpl, pkgName, status)
+			wsHub.Broadcast <- bf.Bytes()
```
kosmoz, line 105: A very similar part occurs three times.
I think it would be OK to refactor this in the future.

internal/web/websocket.go
kosmoz, line 1: I think this file should be renamed to `wsclient.go`

internal/web/hub.go
```diff
@@ -0,0 +1,45 @@
+package web
+
+// Hub maintains the set of active clients and broadcasts messages to the clients.
+type Hub struct {
```
kosmoz, line 4: I think this name is a little confusing. I would rename it to `WsHub`.
(in that case, the file should also be renamed to `wshub.go`

kosmoz (changes requested):
So, this is a pretty big PR but for the most part, I think it looks pretty good. :+1:

There are some things I noticed but it's all just moving stuff around.

pmig (commented):
Wow these updates work great, such a game changer.

As a future improvement we might want to a spinner or some other moving part to the UI indicating that the action is currently in the works and the user doesn't need to perform an action.

internal/web/server.go
kosmoz: OK, let's keep it as is then.

internal/web/server.go
christophenne: Yes, I think also that the handlers need to be pulled out of there and moved to separate files maybe - I think that could be refactored when implementing the package detail page.

kosmoz (approved):

PR #218, opened 2024-02-07
docs: add package reconciliation diagram

kosmoz:
## 📑 Description
Add some documentation about how package updates will be implemented.

## ✅ Checks


pmig (changes requested):
Having multiple files called `index.yaml` is already confusing to me.

I think we should rename the package's `index.yaml` to `versions.yaml`  as it will contain a list of version objects. One of the attribute of this version object could also be the attribute `latest` (boolean or string) so we could also make add pre-releases available in the repository (which would be available via version pinning)

I'm not 100% sure with the new naming. I also thought about calling one of the `index.yaml` files `packages.yaml`, but `versions.yaml` sticked the  most.

pmig (approved):

PR #219, opened 2024-02-08
Changed the properties of img to support svg

### LINKED ISSUE #206: Support svg images via `iconUrl` of package manifest
We needed to fix @thschue contribution for [@keptn](https://github.com/keptn) due to the fact that ui is not yet able to display svg images. The svg is not visible in the UI.

Also support `iconUrl`s that link to an svg.

**Additional context**
Once we fixed this issue we can revert following commit: https://github.com/glasskube/packages/commit/2037ef97a2a655915b11724d548b2f692fd566b4

p1kalys:
Closes #206  

## 📑 Description

Supporting SVG elements in iconUrl

## ✅ Checks

## ℹ Additional Information

Changed the properties of img in packages.html to support svg without changing dimensions. Added a commit to package manifest, iconUrl to particularly accept an url format.

### DIFF
--- a/schema/package-manifest/schema.json
+++ b/schema/package-manifest/schema.json
@@ -63,7 +63,8 @@
           "type": "string"
         },
         "iconUrl": {
-          "type": "string"
+          "type": "string",
+          "format": "uri"
         },
         "helm": {
           "$ref": "#/$defs/HelmManifest"
--- a/schema/package-manifest/schema.json
+++ b/schema/package-manifest/schema.json
@@ -63,8 +63,7 @@
           "type": "string"
         },
         "iconUrl": {
-          "type": "string",
-          "format": "uri"
+          "type": "string"
         },
         "helm": {
           "$ref": "#/$defs/HelmManifest"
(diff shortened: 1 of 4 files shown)


schema/package-manifest/schema.json
```diff
@@ -63,7 +63,8 @@
           "type": "string"
         },
         "iconUrl": {
-          "type": "string"
+          "type": "string",
+          "format": "uri"
```
kosmoz, line 67: The schema file is autogenerated and should not be edited. Unfortunately this isn't well documented yet. 

The source files are in `api/v1alpha1/package_manifest.go` and the schema can be regenerated by calling `make` or `make schema-gen` or, if you don't want to use make, `go run ./internal/cmd/schema-gen -o schema`.

It should be possible to add `jsonschema:"format=uri"` or something like that to the `IconUrl` property of `PackageManifest`. 

This is JSON schema library we use for this: https://github.com/invopop/jsonschema

kosmoz (changes requested):
Hi @p1kalys, thank you for your contribution! :rocket: 

Could you please take a look at the comment below?

schema/package-manifest/schema.json
p1kalys: Thanks for the suggestion, will work on it.

PR #235, opened 2024-02-10
feat: added a feature to detect outdated client for cli (#210)

### LINKED ISSUE #210: Client is outdated detection
Currently there is no hint that a new version of glasskube got released, so users might miss the latest features of Glasskube.

 - An internal version util should be created that is able to fetch the latest version from the hour website.
 - For the GUI there should be an element in nav bar that hints that a new version is available
 - For the CLI we should print a warning that looks something like:
 

>  ```
> $ glasskube ls
>     --------------------------------------------------------------------------------------------------------------- 
>                                                                                                                     
>                                             Update available 0.0.1 → 0.0.2                                          
>                                Please update glasskube to the latest version                                     
>
>     --------------------------------------------------------------------------------------------------------------- 
> 
> NAME                  STATUS         
> cert-manager          Pending          
> cyclops               Not installed    
> ingress-nginx         Not installed    
> keptn                 Ready            
> kubernetes-dashboard  Ready  
> 
> ```

We could host a static file either 

**Additional context**
related to https://github.com/glasskube/glasskube/issues/205
depends on: https://github.com/glasskube/glasskube/pull/209

Baalekshan:
Closes #210

## 📑 Description
Added an internal cliutil - checkforupdate.go which checks for updates and notifies the user

#### printUpdateMessage() - prints the update message in the CLI

Whenever the glasskube is run in the CLI it always checks for the update before executing other subcommands such as list or serve.

image

## ✅ Checks

### DIFF
--- a/internal/cliutils/checkforupdate.go
+++ b/internal/cliutils/checkforupdate.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"regexp"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


internal/cliutils/checkforupdate.go
```diff
@@ -0,0 +1,83 @@
...
+
+	resp, err := http.Get(url)
+	if err != nil {
+		return false, "", fmt.Errorf("failed to fetch release information: %v", err)
+	}
+
+	defer func() {
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			err = fmt.Errorf("failed to close response body: %v", closeErr)
+		}
```
kosmoz, line 47: If I understand this correctly, `err` is assigned here, but it is never used because `err` is never returned in this function. This could work if `err` was part of a [named result](https://go.dev/doc/effective_go#named-results).

In this case though I think it would be preferable to ignore the close error, since we only read from the stream.

To do this without triggering the linter, the result of `Close()` must be explicitly assigned to `_`.

internal/cliutils/checkforupdate.go
```diff
@@ -0,0 +1,83 @@
...
+		return true, releaseInfo.Version, nil
+	}
+
+	return false, "", nil
+}
+
+func UpdateFetch() {
+	updateAvailable, latestVersion, err := checkForUpdate()
+	if err != nil {
+		fmt.Println("Error:", err)
```
kosmoz, line 69: I think it would be better if the update check failed silently.

internal/cliutils/checkforupdate.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+
+	printUpdateMessage := func() {
+		fmt.Printf("\n   --------------------------------------------------------------------------------------------------------------- \n\n")
```
kosmoz, line 74: This should be printed to `os.Stderr`

kosmoz (changes requested):
Hi @Baalekshan, thank you for your contribution. This looks great and works well, but I have some comments, could you please take a look?

Baalekshan:
Hello @kosmoz I will definitely make the changes. Thank you

Baalekshan:
Hi @kosmoz I have made the requested changes, I hope its good now.

pmig:
@Baalekshan don't worry about the failing GitHub action, this will not have an effect on your pr and will not be visible after we rebased this pr.

Baalekshan:
@pmig thank you, but whenever i try to sync my forked branch with the glasskube/main it makes a commit automatically to the pr is there a way to do this silently ?

kosmoz (approved):

PR #251, opened 2024-02-13
feat(ui): add package detail page #172

### LINKED ISSUE #172: Show Package Details on separate page

christophenne:
Closes #172

## 📑 Description
This PR introduces the package detail page, as well as the corresponding changes to the package overview page. The actions available for each package are different on the two views, so I split them into two "components" (template + some controller functions). 
For the new details appearing on the detail page, the PackageManifest schema had to be changed, this is the corresponding PR in the packages repo: https://github.com/glasskube/packages/pull/24 

I also did some refactoring which almost got a little out of hand:
* added mux router (for easier path param handling) and moved all route handlers into separate functions
* moved support page to a separate endpoint
* moved template loading

## ✅ Checks


internal/web/templates/pages/packages.html
```diff
@@ -5,30 +5,32 @@
...
-                <svg width="52" height="52" style="max-width: 52px; max-height: 52px; height: auto;">
-                  <image xlink:href="{{ .IconUrl }}" width="52" height="52" alt="{{ .PackageName}}" />
-                </svg>
-                {{end}}
+                <a class="text-decoration-none" href="/packages/{{ .PackageName }}">
+                  {{if eq .IconUrl ""}}
+                  <!-- TODO the glasskube logo as fallback is probably not the best idea? -->
+                  <img src="/static/assets/glasskube-logo.svg" alt="{{ .PackageName}}"  style="max-width:52px; max-height: 52px; height:auto">
+                  {{else}}
+                  <img src="{{ .IconUrl }}" alt="{{ .PackageName}}"  style="max-width:52px; max-height: 52px; height:auto">
```
christophenne, line 15: I want to have a look at this svg thing again, because it seems to me that it changed the design on the package overview page a bit (in the case of cyclops with the short short description, the logo + text are more horizontally centered instead of at the top). I also have to fix it on the package detail page

christophenne: should be fixed now

internal/web/components/pkg_detail_btns/controller.go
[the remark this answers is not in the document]
christophenne: Yes I agree

kosmoz (approved):

Issue #140, opened 2024-01-30
Extended Kubeconfig Support Page


christophenne:
Follow-Up of #31: The user should be able to drag-and-drop/upload their kubeconfig file if none was found - the UI will copy the given file into the recommended path and re-initializes itself, such that the user doesn't need to do anything anymore. 

The user has two choices: "Only for this session" or "Globally" - that means they can either use the provided kubeconfig only for the current session (i.e. the time the sever process is running), or for all upcoming times until changed (i.e. the kubeconfig file is copied into the recommended home path, possibly affecting other kubernetes-tools too). 

Additional tasks:
- [ ] refactor to not store the `ctx`, see https://github.com/glasskube/glasskube/pull/260#discussion_r1492332245

Depends on #126

PR #260, opened 2024-02-16
feat(ui): bootstrap via UI #123

### LINKED ISSUE #123: Bootstrap via UI
We already have the `glasskube bootstrap` command (#68), but we should also give UI users the possibility to bootstrap glasskube. That means, that the UI has to detect whether it already is bootstrapped, and if not, show an information page + a button to run bootstrap. 

Related #149 #196

Depends on #126

christophenne:
Closes #123

## 📑 Description
This PR introduces a better UI support page for when Glasskube is not bootstrapped in the cluster yet.

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -12,6 +12,12 @@ import (
 	"os"
 	"syscall"
 
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/glasskube/glasskube/internal/config"
+	"github.com/glasskube/glasskube/pkg/bootstrap"
+
 	"github.com/glasskube/glasskube/pkg/manifest"
 
 	"github.com/glasskube/glasskube/pkg/describe"
@@ -51,6 +57,8 @@ type server struct {
 	port       int32
 	listener   net.Listener
 	ctx        context.Context
+	cfg        *rest.Config
+	rawCfg     *api.Config
 	support    *ServerConfigSupport
 	pkgClient  *client.PackageV1Alpha1Client
 	wsHub      *WsHub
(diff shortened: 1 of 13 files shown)


internal/web/server.go
```diff
@@ -237,22 +260,61 @@ func (s *server) packageDetail(w http.ResponseWriter, r *http.Request) {
...
+		err := supportPageTmpl.Execute(w, s.support)
+		checkTmplError(err, "support")
 	} else {
 		http.Redirect(w, r, "/", http.StatusFound)
 	}
 }
 
+func (s *server) bootstrapPage(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		cfg := client.ConfigFromContext(s.ctx)
```
kosmoz, line 281: I think this `cfg` should be stored in the server

internal/web/server.go
```diff
@@ -237,22 +260,61 @@ func (s *server) packageDetail(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			s.support = nil
+			err := bootstrapPageTmpl.ExecuteTemplate(w, "bootstrap-success", nil)
+			checkTmplError(err, "bootstrap-success")
+		}
+	} else {
+		if s.support != nil && s.support.BootstrapMissing {
+			// try to validate bootstrap again, if it failed last time:
+			if s.support.BootstrapCheckError != nil {
+				isBootstrapped, err := bootstrap.IsBootstrapped(s.ctx, client.ConfigFromContext(s.ctx))
```
kosmoz, line 301: use cfg from server (see other comment)

internal/web/templates/pages/bootstrap.html
```diff
@@ -0,0 +1,91 @@
...
+
+{{define "bootstrap-button"}}
+<div class="my-4 d-flex justify-content-center">
+  <button class="btn btn-primary btn-lg py-2" hx-post="/bootstrap" hx-target="#bootstrap-content" hx-disabled-elt="this">
+    {{ . }}
+  </button>
+</div>
+
+<small>
+  This will bootstrap Glasskube in your cluster, by using All-In-One Bootstrapping.
```
kosmoz, line 52: I would rephrase this a little bit:

> This will bootstrap Glasskube in your cluster using an *all-in-one* configuration. If your use-case requires a *slim* configuration or custom manifest, please use the `CLI` command.

internal/web/templates/pages/bootstrap.html
```diff
@@ -0,0 +1,91 @@
+{{define "bootstrap-success"}}
+<div class="col-md-8 p-lg-6 mx-auto text-center">
+  <h1>It's done!</h1>
+</div>
+<div class="col-md-6 p-lg-2 mx-auto">
+  <div class="alert alert-success text-center" role="alert">
+    Congrats, Glasskube is all set up! Have fun managing packages :)
```
kosmoz, line 7: I think I would rather use an emoji for this, e.g. 😊 (`&#128522;&#x1F60A;`)

Also I would add a dot after the sentence

internal/web/server.go
```diff
@@ -237,22 +260,61 @@ func (s *server) packageDetail(w http.ResponseWriter, r *http.Request) {
...
+		checkTmplError(err, "support")
 	} else {
 		http.Redirect(w, r, "/", http.StatusFound)
 	}
 }
 
+func (s *server) bootstrapPage(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		cfg := client.ConfigFromContext(s.ctx)
+		client := bootstrap2.NewBootstrapClient(
```
kosmoz, line 282: I think it would be ok to move the IsBootstrapped function from internal to pkg. Then we would not need the two imports and could just use bootstrap.

If you prefer to keep it as is, I think the import name should be changed to something like `internalBootstrap`

christophenne:
> Also I think it might be a good idea to show the current context name or cluster name on the bootstrap page so a user knows which cluster is going to be bootstrapped. What do you think?

Yes I think that's a good idea, I will add that. 
In the long term I think it also makes sense to show that kind of information somewhere else too, maybe in the navbar together with a status indicator, or on a separate page with all kinds of system information? Because one could manage multiple clusters and start multiple UIs for that, which could become very confusing.

internal/web/server.go
```diff
@@ -87,6 +97,20 @@ func (s *server) Start(ctx context.Context, support *ServerConfigSupport) error
...
+				BootstrapCheckError: err,
+			}
+		}
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "\nFailed to check whether Glasskube is bootstrapped: %v\n\n", err)
+		}
+	}
+
 	s.support = support
 	s.ctx = ctx
```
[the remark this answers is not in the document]
christophenne: Ok interesting, I haven't really thought about the contexts yet. Maybe I'll change it right in this PR.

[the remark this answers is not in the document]
christophenne: Now I see what you mean, I would repair/refactor the context storing in #140 because there we will need to reload the config and reinitialize some things, I think that could be a good fit.

kosmoz (approved):

PR #261, opened 2024-02-16
chore: change schema to not use $ref and move to website

### LINKED ISSUE #258: Expose schema files via our website
Currently the generated schema files must be copied manually between repositories. 

We should make them publicly available via our website: e.g. `https://glasskube.dev/schema(s)?/package/schema.json` 

**Additional context**
Slightly related to: https://github.com/glasskube/glasskube/pull/256

kosmoz:
Closes #258 

## 📑 Description
With this PR, our schema files no longer use $ref, since it is strictly speaking not standards compliant and not very well supported by some tooling. 

The schema files are now written to website/static, which makes them available via our website. 

The $id of the schema reflects this change.

## ✅ Checks

## ℹ Additional Information

Once this is merged, we can remove the schema files from the packages repo and update all packages to reference the new schema location

### DIFF
--- a/.github/workflows/schema.yaml
+++ b/.github/workflows/schema.yaml
@@ -5,12 +5,12 @@ on:
     branches:
       - main
     paths:
-      - schema/**
+      - website/static/schemas/**
   pull_request:
     branches:
       - main
     paths:
-      - schema/**
+      - website/static/schemas/**
 
 jobs:
   validate:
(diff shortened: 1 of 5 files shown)


pmig (commented):
Would it be possible, to still use $ref, but just not in the root of the package manifest?

This change results in different type generations. Using refs we get cleaner interfaces:

image

kosmoz:
> Would it be possible, to still use $ref, but just not in the root of the package manifest?

Not with the tools we use at the moment. It's all or nothing

pmig (approved):

PR #265, opened 2024-02-17
feat: add version command

### LINKED ISSUE #254: Add `version` command to glasskube
A version command comes in handy to get the version of all glasskube components

Similar to [flux version command](https://github.com/fluxcd/flux2/blob/main/cmd/flux/version.go#L157):

```shell
$ flux version
flux: v2.2.3
distribution: flux-v2.2.3
helm-controller: v0.37.4
kustomize-controller: v1.2.2
source-controller: v1.2.4
```

A glasskube version command could look like

```shell
$ glasskube version
glasskube: v0.0.3
package-operator: v0.0.2
```

If the glasskube deployment could not be found an error message should be printed:

> ✗ no deployments found in the glasskube-system namespace

**Additional context**
As linked above flux parses image name and tag and prints it in a list. We could to the same which would result in the needed result.  Also the `glasskube -v` command does not need to change.

This issue is related to #210 and can help #205

Baalekshan:
Closes #254 

## 📑 Description
Added a version command which prints the version of glasskube and package-operator.
```
$ glasskube version
glasskube: v0.0.3
package-operator: v0.0.2 
```
If no deployments are found under glasskube-system namespace it prints
```
$ glasskube version
glasskube: v0.0.3
✗ no deployments found in the glasskube-system namespace
```

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/version.go
+++ b/cmd/glasskube/cmd/version.go
@@ -35,7 +35,7 @@ func init() {
 }
 
 func getPackageOperatorVersion(ctx context.Context) (string, error) {
-	config := cliutils.RequireConfig("")
+	config, _ := cliutils.RequireConfig("")
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		return "", err
--- a/cmd/glasskube/cmd/version.go
+++ b/cmd/glasskube/cmd/version.go
@@ -1,0 +1,77 @@
+package cmd
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/glasskube/glasskube/internal/config"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+var versioncmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print the version of glasskube and package-operator",
+	Long:  `Print the version of glasskube and package-operator`,
+	Run: func(cmd *cobra.Command, args []string) {
+		glasskubeVersion := config.Version
+		fmt.Fprintf(os.Stderr, "glasskube: v%s\n", glasskubeVersion)
+		operatorVersion, err := getPackageOperatorVersion()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "✗ no deployments found in the glasskube-system namespace\n")
+		} else {
+			fmt.Fprintf(os.Stderr, "package-operator: %s\n", operatorVersion)
+		}
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(versioncmd)
+}
+
+func getPackageOperatorVersion() (string, error) {
+	var kubeconfig string
+	if home := homedir.HomeDir(); home != "" {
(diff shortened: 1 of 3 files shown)


cmd/glasskube/cmd/version.go
```diff
@@ -0,0 +1,77 @@
...
+		flag.StringVar(&kubeconfig, "kubeconfig", home+"/.kube/config", "(optional) absolute path to the kubeconfig file")
+	} else {
+		flag.StringVar(&kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file")
+	}
+	flag.Parse()
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return "", err
+	}
```
kosmoz, line 51: We already have a `PreRun` util for this, which can be used by adding `PreRun: cliutils.SetupClientContext(true)` to the cmd.

The kubeconfig flag already exists on the root cmd, so you don't have to implement your own.

cmd/glasskube/cmd/version.go
```diff
@@ -0,0 +1,77 @@
...
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return "", err
+	}
+
+	namespace := "glasskube-system"
+	deploymentName := "glasskube-controller-manager"
+	deployment, err := clientset.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
```
kosmoz, line 60: Please use a "real" context here. You can use `cmd.Context()` in the `Run` func to retrieve the cmd context.

cmd/glasskube/cmd/version.go
```diff
@@ -0,0 +1,77 @@
...
+
+	containers := deployment.Spec.Template.Spec.Containers
+	for _, container := range containers {
+		if container.Name == "manager" {
+			ref, err := name.ParseReference(container.Image)
+			if err != nil {
+				return "", err
+			}
+			parts := strings.Split(ref.Identifier(), ":")
+			return parts[0], nil
```
kosmoz, line 73: I think this is unnecessary. `ref.Identifier()` would be enough.

kosmoz (changes requested):
Hi @Baalekshan, thanks for you PR! :rocket: 

I tried the new feature and it works well, but I do have some comments. Could you please take a look?

cmd/glasskube/cmd/version.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+var versioncmd = &cobra.Command{
+	Use:    "version",
+	Short:  "Print the version of glasskube and package-operator",
+	Long:   `Print the version of glasskube and package-operator`,
+	PreRun: cliutils.SetupClientContext(true),
```
kosmoz, line 20: The version cmd should also work in a non-bootstrapped cluster, but this is prevented here. To change this, you can change this to `cliutils.SetupClientContext(false)`

cmd/glasskube/cmd/version.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(versioncmd)
+}
+
+func getPackageOperatorVersion(ctx context.Context) (string, error) {
+	config, _ := cliutils.RequireConfig("")
```
kosmoz, line 38: Please use `client.ConfigFromContext(ctx)` to get the config from the context. See github.com/glasskube/glasskube/pkg/client

kosmoz (changes requested):
This looks pretty good now. I found two more issues, after that we should be good to go :-)

kosmoz (approved):

Issue #168, opened 2024-02-01
add progress bar and spinner in all CLI commands


tommuhm:
the bootstrap command has nice visual effects in the terminal, it would be good to do something similar for install/uninstall comands to achiever higher consistency amongst the various commands

kosmoz:
This is already implemented for the `install` cmd. 

To anyone interested in giving this a go, be sure to review `pkg/install/install.go` for reference before you start, so we can avoid duplicate code!

christophenne:
related to #133 - both issues need to introduce a blocking `Uninstall` method, ideally this should only be done once.

fykaa:
If nobody is currently working on this, would like to get assigned!

christophenne:
@fykaa sure, go ahead :) 

As mentioned, in #133 the same blocking uninstallation will need to be used

fykaa:
Hey @christophenne ! I'm doing well, thank you! 

I'm currently in the process of gathering all the necessary information and resources to do this effectively. I haven't started the actual work yet, but rest assured, I'll try to make a headway on it soon. I'm sorry for any hold-ups caused by this Issue, I'll keep you posted on my progress right here

PR #272, opened 2024-02-19
feat: add blocking owner deletion 

kosmoz:
## 📑 Description
With this PR, foreground deletion is enabled when uninstalling a Package.

## ✅ Checks

## ℹ Additional Information

Note: The uninstall cmd still doesn't wait for the resource to be deleted, but the functionality can be tested using `kubectl delete --cascade=foreground pkg <package-name>`.

* Relevant documentation:
   * https://kubernetes.io/docs/concepts/architecture/garbage-collection/#foreground-deletion
   * https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#use-foreground-cascading-deletion
* This unblocks #133 and #168

### DIFF
--- a/pkg/client/package.go
+++ b/pkg/client/package.go
@@ -17,7 +17,7 @@ type PackageInterface interface {
 	Get(ctx context.Context, pkgName string, p *v1alpha1.Package) error
 	GetAll(ctx context.Context, result *v1alpha1.PackageList) error
 	Watch(ctx context.Context) (watch.Interface, error)
-	Delete(ctx context.Context, pkg *v1alpha1.Package) error
+	Delete(ctx context.Context, pkg *v1alpha1.Package, options metav1.DeleteOptions) error
 }
 
 type packageClient struct {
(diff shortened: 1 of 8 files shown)


pkg/client/package.go
```diff
@@ -51,10 +51,16 @@ func (c *packageClient) GetAll(ctx context.Context, result *v1alpha1.PackageList
 		Do(ctx).Into(result)
 }
 
-func (c *packageClient) Delete(ctx context.Context, pkg *v1alpha1.Package) error {
+func (c *packageClient) Delete(ctx context.Context, pkg *v1alpha1.Package, blocking bool) error {
```
christophenne, line 54: Not sure about the flag argument here, I would prefer two separate methods or to pass the propagation as parameter (since the client is a bit more "low-level" I could imagine it makes sense to expose this detail to the caller).

christophenne: Yes that's a good idea.

christophenne (approved):

PR #275, opened 2024-02-20
feat(cli): add showing package version in list cmd (#200)

### LINKED ISSUE #200: Add a column for update status in `glasskube list` output

kosmoz:
Closes #200 

## 📑 Description
This PR adds two new columns to the output of `glasskube list`:
* VERSION: shows the installed version plus the spec version and/or available version (from the registry) if they are different
* LATEST VERSION: shows the latest version (from the registry) - Disabled by default

`glasskube list` now has two additional flags: 
* `--show-latest`: enables the LATEST VERSION column (see above)
* `--more` or `-m`: equivalent to `--show-description --show-latest`

## ✅ Checks

## ℹ Additional Information
I also took the liberty and moved all command specific config variables to their respective command files, as there is no need for a global variable for those.

Some parts of the version column rely on a `latestVersion` being present for the package in the package index (see updated schema file). This is currently not the case and should be done automatically in the future but for testing you will have to set the `latestVersion` yourself in the index.yaml file.

### DIFF
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -3,13 +3,20 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/glasskube/glasskube/internal/cliutils"
-	"github.com/glasskube/glasskube/internal/config"
 	"github.com/glasskube/glasskube/pkg/client"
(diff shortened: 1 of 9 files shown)


kosmoz:
Note: I will update the docs after the review

christophenne (approved):
LGTM, I'm also fine with the new options

PR #277, opened 2024-02-20
feat(cli): add installing specific package version (#203)

### LINKED ISSUE #203: Extend `glasskube install` to support installing a particular version of a package ( Version pinning )
There are multiple ways this could be done:
 - `glasskube install package@v1.0.0`
 - `glasskube install package --version=v1.0.0`
 - … something else?

We need to decide. Input welcome!

When no version i specified, the latest version should be installed.

kosmoz:
Closes #203  

## 📑 Description
This PR adds two additional flags to the `glasskube install` command:
* `--version`: install a specific version of a package
* `--enable-auto-updates`: always keep this package at the latest version

If neither is specified, the user is asked if automatic updates should be enabled. 
If they answer "no", the latest version of the package is fetched and used for installation.

## ✅ Checks

## ℹ Additional Information
Web UI is still TODO.

~Docs will be updated after review.~ Docs updated

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -14,6 +14,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var installCmdOptions = struct {
+	Version           string
+	EnableAutoUpdates bool
+}{}
+
 var installCmd = &cobra.Command{
 	Use:               "install [package-name]",
 	Short:             "Install a package",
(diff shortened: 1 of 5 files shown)


cmd/glasskube/cmd/install.go
```diff
@@ -68,6 +93,34 @@ func completeAvailablePackageNames(
...
+	for _, version := range packageIndex.Versions {
+		if toComplete == "" || strings.HasPrefix(version.Version, toComplete) {
+			versions = append(versions, version.Version)
+		}
+	}
+	return versions, 0
+}
+
 func init() {
+	installCmd.PersistentFlags().StringVar(&installCmdOptions.Version, "version", "",
```
christophenne, line 119: I think a `-v` shorthand would be nice here

christophenne (approved):

pmig (approved):

PR #282, opened 2024-02-21
feat(ui): add selecting kubeconfig (#140)

### LINKED ISSUE #140: Extended Kubeconfig Support Page

kosmoz:
Closes #140 

## 📑 Description
TODO

## ✅ Checks

### DIFF
--- a/internal/web/templates/pages/support.html
+++ b/internal/web/templates/pages/support.html
@@ -1,40 +1,86 @@
 {{define "body"}}
-<div class="position-relative overflow-hidden p-3 p-md-5 m-md-3">
-  <div class="col-md-8 p-lg-8 mx-auto text-center">
-    <h1>You're almost there!</h1>
-  </div>
+<div class="container">
+  <div class="row">
+    <div class="col-md-8 offset-md-2 col-lg-6 offset-lg-3">
+      <div class="text-center my-3">
+        <h1>You're almost there!</h1>
+      </div>
 
-  {{if .KubeconfigMissing}}
-  <div class="col-md-6 p-lg-4 mx-auto">
-    <p>
-      To use Glasskube, please download the kubeconfig file from your cloud provider and copy it to this location on your computer:
-    </p>
-    <p class="text-center lead mb-3 font-monospace">{{ .KubeconfigDefaultLocation }}</p>
-    <p>
-      Afterwards, stop the Glasskube Server in your terminal with CTRL+C, and then restart it with <code>glasskube serve</code>.
-    </p>
-    <hr>
-    <small>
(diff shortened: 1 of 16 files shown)


christophenne:
I know it's a draft, but before I forget to mention it: it only partially closes #140

kosmoz:
> I know it's a draft, but before I forget to mention it: it only partially closes #140

Thanks, but yes, that's why it's still a draft :stuck_out_tongue_closed_eyes:

internal/web/templates/pages/support.html
```diff
@@ -1,40 +1,86 @@
...
+      </small>
+    </div>
+  </div>
 </div>
+
+<script type="text/javascript">
+  /**
+   * @type {HTMLInputElement}
+   */
+  const kubeconfigInput = document.getElementById('kubeconfig');
```
christophenne, line 60: During testing I encountered a bug (that I now can't reproduce anymore), but I think it was related to this line. When you upload an invalid kubeconfig, then "Go Back" to the support page, the file upload does not work anymore. Browser console says "Uncaught SyntaxError: Identifier 'kubeconfigInput' has already been declared".
Unfortunately I don't know exactly enough how I got to that point.

christophenne: Ok I think it was because I went back with the browser back button instead of the "Go back" link, in that case the upload doesn't work anymore (however without the browser console error). I think that's a marginal issue, the case of an invalid kubeconfig file will not happend that often and if so, people probably press the "Go back" button.

christophenne (changes requested):
Lots of good changes I think, I just have a couple smaller (and one bigger) comments, see comments.

internal/web/templates/pages/support.html
kosmoz: Interesting. I think the problem could be that the JS is executed twice due to HTMX swapping the page instead of a full page reload. I'll check and see what I can do.

kosmoz: I am unable to reproduce this bug in both Chromium and Firefox :confused:

christophenne (approved):
Nice! :rocket:

PR #286, opened 2024-02-23
feat(ui): update packages and push cluster events to UI (#269)

### LINKED ISSUE #269: Support package version handling in UI
Users cannot install a praticular version of a package via UI yet.

* On click on the install button, a modal should appear, where the user can select whether they want to have the latest version and auto-updates (default enabled), or a specific version. This should happen on both pages where the Install button is present
* When a package is installed in a pinned version, and this version is outdated, the user should see that their version is outdated and have the possibility to update their package too. 

Once we have configuration in place, we need a separate page anyway, but for now an extra page would look too empty and unnecessary. 

**Additional context**
Latest version of a package might need to be persisted in the package repo `index.yaml` – otherwise we would have to select the `versions.yaml` for each package. Probably scout will handle the repo update automatically.

christophenne:
Closes #269 

## 📑 Description
TODO

## ✅ Checks


internal/web/server.go
```diff
@@ -216,29 +206,64 @@ func (s *server) installModalVersions(w http.ResponseWriter, r *http.Request) {
...
+		})
+		checkTmplError(err, "pkgUpdateModalTmpl")
+	}
+}
+
+func (s *server) update(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	pkgName := r.FormValue("packageName")
+
+	updater := update.NewUpdater(s.pkgClient).WithStatusWriter(statuswriter.Stderr())
```
kosmoz, line 239: The updater is actually designed to be reusable, so I think it would be better to have it as a server property.

internal/web/server.go
```diff
@@ -216,29 +206,64 @@ func (s *server) installModalVersions(w http.ResponseWriter, r *http.Request) {
...
+	ut, err := updater.Prepare(ctx, []string{pkgName})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "An error occurred preparing update of %v: \n%v\n", pkgName, err)
+		return
+	}
+	// in the future we might want to check here whether the prepared new version is the same is the "toVersion"
```
kosmoz, line 245: Typo: "is the same ~is~as" :slightly_smiling_face:

internal/web/server.go
```diff
@@ -468,6 +498,49 @@ func defaultKubeconfigExists() bool {
...
+				return &pkgList, err
+			},
+			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+				return s.pkgClient.Packages().Watch(context.Background())
+			},
+		},
+		&v1alpha1.Package{},
+		0,
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
```
kosmoz, line 516: I think it is a little more modern to use `any` instead of `interface{}`. It's just an alias.

internal/web/server.go
```diff
@@ -468,6 +498,49 @@ func defaultKubeconfigExists() bool {
...
+					s.broadcastPkg(pkg, client.GetStatusOrPending(&pkg.Status), nil, "")
+				}
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				if pkg, ok := newObj.(*v1alpha1.Package); ok {
+					ctx := client.SetupContextWithClient(context.TODO(), s.restConfig, s.rawConfig, s.pkgClient)
+					if mf, err := manifest.GetInstalledManifestForPackage(ctx, *pkg); err != nil && !errors.Is(err, manifest.ErrPackageNoManifest) {
+						fmt.Fprintf(os.Stderr, "Error fetching manifest for package %v: %v\n", pkg.Name, err)
+					} else {
+						var idx repo.PackageIndex
```
kosmoz, line 527: Since the `latestVersion` is now in the repo index as well, I think it would make sense to download that once and then keep it in memory. That way, we don't have to make a new request to the repo every time a package is updated (which can happen quite frequently).

internal/web/server.go
```diff
@@ -468,6 +498,49 @@ func defaultKubeconfigExists() bool {
 	}
 }
 
+func (s *server) initInformer() (cache.Store, cache.Controller) {
+	return cache.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+				var pkgList v1alpha1.PackageList
+				err := s.pkgClient.Packages().GetAll(context.Background(), &pkgList)
```
kosmoz, line 506: I think it would be good to use the `cmd.Context()` here, which was removed from the `Start` function last week :upside_down_face:

kosmoz (changes requested):
Overall, I really like these changes! Looks very good, although I do have some small comments. 

Also, via the CLI it is possible to update all outdated packages at once. This is not yet possible via the UI. Should we create a separate ticket for this feature?

christophenne:
> Overall, I really like these changes! Looks very good, although I do have some small comments.
> 
> Also, via the CLI it is possible to update all outdated packages at once. This is not yet possible via the UI. Should we create a separate ticket for this feature?

I created #289

internal/web/server.go
christophenne: I implemented a mini-cache which holds the index for 5 minutes now, but I don't like it one bit and I think we should migrate it to a proper cache for all the repo methods. What do you think?

kosmoz: I agree that the current implementation seems a little half baked. What do you have in mind regarding a "proper" cache?

christophenne: I meant to say that we should consider caching at other places too (all sorts of manifest and repo index fetching), probably also in the operator? 
Something like this library: https://github.com/eko/gocache could help.

christophenne: I will create a ticket, I think its out of scope for this one.

internal/web/server.go
christophenne: As discussed, we will leave it as it is for now, because we don't know how the status writer handles concurrent usage.

PR #288, opened 2024-02-26
feat(cli): change update message to be less obtrusive

kosmoz:
## 📑 Description
This PR includes several changes to the already existing update check feature:
* It is now less obtrusive and more similar to our other CLI messages
* There is a new (global) `--skip-update-check` flag that does what it says it does
* `--skip-update-check` is `true` by default in development builds (this can be overridden by using `--skip-update-check=false`

## ✅ Checks

## ℹ Additional Information
I did not update the documentation, I'm not sure if it's even necessary to document the new flag outside of the `help` command…

### DIFF
--- a/cmd/glasskube/cmd/root.go
+++ b/cmd/glasskube/cmd/root.go
@@ -3,21 +3,33 @@ package cmd
 import (
 	"fmt"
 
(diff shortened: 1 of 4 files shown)


christophenne (approved):
Great idea :rocket:

PR #290, opened 2024-02-26
feat(cli): add outdated flag for list cmd (#201)

### LINKED ISSUE #201: Add `--outdated` flag to `glasskube list` to only shows outdated packages

kosmoz:
Closes #201 

## 📑 Description
This PR adds a a new flag `--outdated` to the list subcommand. This flag is false by default. If it is set to true, the output is restricted to those packages with an outdated version.

## ✅ Checks

## ℹ Additional Information
I also refactored the ListOptions type from an int type to a struct because I think it is much more convenient and less error prone.

### DIFF
--- a/pkg/list/list.go
+++ b/pkg/list/list.go
@@ -18,32 +18,17 @@ type PackageWithStatus struct {
 	InstalledManifest *v1alpha1.PackageManifest
 }
 
-type listOptions int64
-
-const (
-	IncludePackageInfos listOptions = 1 << iota
-	OnlyInstalled
-)
-
-const (
-	DefaultListOptions listOptions = 0
-)
-
-func Get(client *client.PackageV1Alpha1Client, ctx context.Context, name string) (*v1alpha1.Package, error) {
-	var pkg v1alpha1.Package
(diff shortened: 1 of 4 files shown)


pmig (approved):

pkg/list/list.go
```diff
@@ -56,7 +41,10 @@ func GetPackagesWithStatus(
 			PackageRepoIndexItem: *item.IndexItem,
 		}
 
-		if !onlyInstalled || item.Package != nil {
+		if (!options.OnlyInstalled || item.Package != nil) &&
+			(!options.OnlyOutdated || (item.Package != nil && item.IndexItem != nil &&
+				item.Package.Spec.PackageInfo.Version != item.IndexItem.LatestVersion)) {
```
christophenne, line 46: I think this inner condition lacks another `&&`: `item.Package.Spec.PackageInfo.Version != ""`, otherwise the auto-update-enabled packages will always appear in the list of outdated packages. 

The whole `if` is hard to read in my opinion, it might make sense to store parts of it in bool variables or use functions.

christophenne (changes requested):

christophenne (approved):

PR #293, opened 2024-02-27
feat(ui): update all packages #289

### LINKED ISSUE #289: Update-All via UI
The `update` command in the CLI supports updating all packages at once, but the UI does not.

Support Update-All via UI.

**Additional context**
Related #269

christophenne:
Closes #289

## 📑 Description
Added an Update All button on the package overview page if there is at least one package not up to date. 

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -216,29 +216,45 @@ func (s *server) installModalVersions(w http.ResponseWriter, r *http.Request) {
 
 func (s *server) updateModal(w http.ResponseWriter, r *http.Request) {
 	pkgName := r.FormValue("packageName")
-	var pkg v1alpha1.Package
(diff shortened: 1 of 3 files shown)


kosmoz (approved):
Works great. The only thing I would personally change would be moving the "Update all" button to the right edge of the alert, but I think it would be OK to leave it as well.

kosmoz:
I also created https://github.com/glasskube/glasskube/issues/295 as a follow up to address the comment you added in your previous PR regarding to check

> "whether the prepared new version is the same as the "toVersion"

christophenne:
> Works great. The only thing I would personally change would be moving the "Update all" button to the right edge of the alert, but I think it would be OK to leave it as well.

Yes it looks better, I changed it.

PR #294, opened 2024-02-27
feat(cli): glasskube describe (#241)

### LINKED ISSUE #241: `glasskube describe` command
We are working on the package detail page on the UI (#172 ), so there should be an equivalent command for the CLI to show all relevant package details. Just like on the UI, this should work for both uninstalled and installed packages, and the status of the package should be shown. 

`glasskube describe [package-name]` should print the same package details to the console as we show in the UI. 

**Additional context**
Depends on #172 to develop the client functionality that can be used by both layers. 

For implementing this task, reuse `DescribePackage` from `pkg/describe/describe.go`, which contains all the logic to fetch all the relevant data which should be displayed.

christophenne:
Closes #241

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/describe.go
+++ b/cmd/glasskube/cmd/describe.go
@@ -1,0 +1,70 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/glasskube/glasskube/internal/repo"
+
+	"github.com/glasskube/glasskube/internal/cliutils"
+	"github.com/glasskube/glasskube/pkg/client"
+	"github.com/glasskube/glasskube/pkg/describe"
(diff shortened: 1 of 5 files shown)


christophenne:
> Works great and the code changes look clean.
> 
> I have some comments (mostly UX) and some other thoughts (which I think should be implemented but which we can create separate tasks for):
> 
> * The installed version should be displayed (also on the UI).
> * If there is an endpoint in the manifest, the describe command should tell the user that they can use `glasskube open ...` for this package.
> * The describe command could also work with no kubeconfig, showing a warning at the top and then simply downloading the manifest from  source.
> * We could display the icon, converted to ASCII art. (This is a joke, of course, but it would be kind of funny)

These are all good points, I created #296, #297, #298.

kosmoz (approved):

PR #313, opened 2024-02-28
feat(ui): add dark mode

kosmoz:
## 📑 Description
Let's go dark :sunglasses: 

## ✅ Checks


christophenne (approved):
Looks and works great!

The only thing I'm not so sure about is how the modals look in dark mode. I think it's difficult to grasp that the modal is "on top" of the page – or at least it's easier to see in light mode that something is on top.

kosmoz:
> Looks and works great!
> 
> The only thing I'm not so sure about is how the modals look in dark mode. I think it's difficult to grasp that the modal is "on top" of the page – or at least it's easier to see in light mode that something is on top.

This is true. I will update the theme and re-import it.

Makefile
```diff
@@ -97,7 +97,7 @@ build: manifests generate fmt vet lint ## Build manager binary.
 	go build -o $(OUT_DIR)/package-operator ./cmd/package-operator/
 
 .PHONY: build-cli
-build-cli: fmt vet lint ## Build cli binary.
+build-cli: fmt vet lint lint-web web ## Build cli binary.
```
christophenne, line 100: I think the `lint` target should now be called `lint-go`, or if it stays `lint`, it should also execute `lint-web`. But in the latter the `lint-fix` would also need to include a new target `lint-fix-web` (which itself would probably execute `prettier:write`?).

christophenne (changes requested):
Nice changes :rocket: 

The only other comment I have, is that a short section in the CONTRIBUTING.md would be nice: how to set up for web development or something like that – since now one needs to run `npm i`.

christophenne (approved):

Issue #117, opened 2024-01-30
Uninstall dependent packages


christophenne:
When uninstalling a package which has dependent packages, the latter have to be uninstalled as well.

christophenne:
closed by #314

PR #314, opened 2024-02-28
feat(package-controller): install package dependencies (#111)

### LINKED ISSUE #111: Install packages with dependencies
This needs further specification, but the idea is that every package can have other packages as dependencies. When installing such a package, the dependencies need to be installed too.

- [ ] Adapt package manifest to support dependencies
- [ ] Implement installation

christophenne:
Closes #111 #117

## 📑 Description
This PR introduces the concept of dependencies in packages. This is the most basic implementation without version support or any specific support on CLI/UI.

Documentation about the whole topic is intended to be done in #316 

## ✅ Checks


kosmoz (changes requested):
Looks cool. 

One think I noticed during testing is that it sometimes takes a while to reconcile a package if its dependencies become ready. I think this could be improved by adding an `Owns` call with `Package` in the `SetupWithManager` function. In theory this should tell the controller manager to reconcile a package if another package it owns has changed.

PR #320, opened 2024-02-29
feat(cli): add --no-await in install command

### LINKED ISSUE #133: Add CLI-Flag to not await install/uninstall commands

hanshal101:
Fixes #133 

## 📑 Description
Add CLI-Flag to not await install/uninstall commands

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -65,35 +65,26 @@ var installCmd = &cobra.Command{
 			cancel()
 		}
 
-		// Non-blocking install if --no-wait is used
 		if installCmdOptions.NoWait {
-			go func() {
-				if err := installer.Install(ctx, packageName, installCmdOptions.Version); err != nil {
-					fmt.Fprintf(os.Stderr, "An error occurred during installation:\n\n%v\n", err)
-					os.Exit(1)
-				}
-				fmt.Printf("Installation of %v started in the background.\n", packageName)
-			}()
-			return
-		}
-
-		// Blocking install
-		status, err := installer.InstallBlocking(ctx, packageName, installCmdOptions.Version)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "An error occurred during installation:\n\n%v\n", err)
-			os.Exit(1)
-		}
-		if status != nil {
-			switch status.Status {
-			case string(condition.Ready):
-				fmt.Printf("✅ %v is now installed in %v.\n", packageName, config.CurrentContext)
-			default:
-				fmt.Printf("❌ %v installation has status %v, reason: %v\nMessage: %v\n",
-					packageName, status.Status, status.Reason, status.Message)
-			}
+			installer.Install(ctx, packageName, installCmdOptions.Version)
 		} else {
-			fmt.Fprintln(os.Stderr, "Installation status unknown - no error and no status have been observed (this is a bug).")
-			os.Exit(1)
+			status, err := installer.InstallBlocking(ctx, packageName, installCmdOptions.Version)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "An error occurred during installation:\n\n%v\n", err)
+				os.Exit(1)
+			}
+			if status != nil {
+				switch status.Status {
+				case string(condition.Ready):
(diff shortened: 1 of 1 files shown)


cmd/glasskube/cmd/install.go
```diff
@@ -61,6 +62,34 @@ var installCmd = &cobra.Command{
 			cancel()
 		}
 
+		// if --no-wait is used
+		if installCmdOptions.NoWait {
+			go func() {
+				status, err := install.NewInstaller(client).
+					WithStatusWriter(statuswriter.Spinner()).
+					InstallBlocking(ctx, packageName, installCmdOptions.Version)
```
kosmoz, line 70: The installer instantiation can be lifted out of the "if" statement to reuse some code. 

Something like this:
```go
installer := install.NewInstaller(client).WithStatusWriter(statuswriter.Spinner())
if installCmdOptions.NoWait {
    ...
} else {
    ...
```

kosmoz, line 70: The `installer` already provides a non-blocking variant of the `Install` action. You can simply call `installer.Install(...)` instead of `installer.InstallBlocking(...)`. This removes the need for a goroutine here.

cmd/glasskube/cmd/install.go
```diff
@@ -61,6 +62,34 @@ var installCmd = &cobra.Command{
...
+					case string(condition.Ready):
+						fmt.Printf("✅ %v is now installed in %v.\n", packageName, config.CurrentContext)
+					default:
+						fmt.Printf("❌ %v installation has status %v, reason: %v\nMessage: %v\n",
+							packageName, status.Status, status.Reason, status.Message)
+					}
+				} else {
+					fmt.Fprintln(os.Stderr, "Installation status unknown - no error and no status have been observed (this is a bug).")
+					os.Exit(1)
+				}
```
kosmoz, line 87: The non-blocking `Install` action (see other comment) does not return a status value, obviously, but we don't want to handle that anyways  if --no-wait has been used in the call.

kosmoz (changes requested):
Hi @hanshal101, thank you so much for your second contribution!

I checked out your changes and noticed some improvements. Could you please take a look?

hanshal101:
Hey @kosmoz I have made the necessary changes here (latest commit). Please do share your feedback.

kosmoz:
> Hey @kosmoz I have made the necessary changes here (latest commit). Please do share your feedback. Also I found some error in checks please guide me in resolving those errors.

The Install function returns an error. You have to check it.

hanshal101:
Hey @kosmoz is this is possible solution:
```
if err := installer.Install(ctx, packageName, installCmdOptions.Version); err != nil {
	fmt.Fprintf(os.Stderr, "An error occurred during installation:\n\n%v\n", err)
	os.Exit(1)
}
```

kosmoz:
> Hey @kosmoz is this is possible solution:
> 
> ```
> if err := installer.Install(ctx, packageName, installCmdOptions.Version); err != nil {
> 	fmt.Fprintf(os.Stderr, "An error occurred during installation:\n\n%v\n", err)
> 	os.Exit(1)
> }
> ```

Yes, that should work.

hanshal101:
Made all changes.

hanshal101:
Also should add these same changes in the uninstall command.

kosmoz:
> Also should add these same changes in the uninstall command.

Yes, do you want to do it in the same PR?

hanshal101:
> > Also should add these same changes in the uninstall command.
> 
> Yes, do you want to do it in the same PR?

You say what's better.

kosmoz:
> > > Also should add these same changes in the uninstall command.
> > 
> > 
> > Yes, do you want to do it in the same PR?
> 
> You say what's better.

Let's use a separate PR then. :slightly_smiling_face:

hanshal101:
I think you should merge this PR first then I will be able to make another PR from different branch

kosmoz (approved):

hanshal101:
Hey @kosmoz will push the other PR by tomorrow. Sorry for the delay.

hanshal101:
Heyy @kosmoz in the uninstall I wonder if we need go routine for the no-wait command.

PR #321, opened 2024-03-01
build: use glasskube/theme dependency from npmjs

christophenne:
## 📑 Description
Previously the build failed, because the package-lock.json file directed the new @glasskube/theme dependency to the "private" github package instead of the npmjs package.

The change has been done by removing the private auth token from my local .npmrc file, removing node_modules and reinstall dependencies with `npm i` – npm then uses the npmjs package instead of the github one. Probably there would have been a better way but this seems to work. However, if one changes back to have the auth token in the .npmrc file, the Github registry is used preferrably by npm.

For the future, I think it would be better to not publish the theme to the github registry at all to avoid this.

## ✅ Checks

## ℹ Additional Information
Related #313

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "glasskube-pm",
+  "name": "glasskube",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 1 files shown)


kosmoz (approved):

kosmoz:
Thanks, yeah publishing to both is probably not a good idea. 

This was only done for the current v0.1.1, since there haven't been any changes to the library. Future versions will only be published to NPM.

PR #326, opened 2024-03-04
fix: change opener to choose first ready pod

### LINKED ISSUE #302: Package `open` can fail if target pod is not yet ready to accept connections
**Describe the bug**
We don't check if a pod is ready to accept connections before we initiate the port-forward.

**To Reproduce**
1. Install a package that has an entry-point (e.g. kubernetes-dashboard)
2. Immediately click on the Open button. 
3. See error

**Expected behavior**
The port-forward should only be initiated if the pod is ready and wait until the pod gets ready for some time (e.g. 5 seconds), else the port-forward should not be initiated and will be initiated by the next attempt.

**Screenshots**

 - Type: Kubernetes
 - Version: v1.28.3
 - Provider: minikube

 - Package Operator version: v0.0.3
 - CLI version: v0.0.3

**Additional context**

Command line logs:

```
Installing kubernetes-dashboard...
	 |I| Forwarding from 127.0.0.1:8443 -> 8443
	 |I| Forwarding from [::1]:8443 -> 8443
	 |I| Handling connection for 8443
E0227 16:13:58.327794  511974 portforward.go:409] an error occurred forwarding 8443 -> 8443: error forwarding port 8443 to pod 19fc729545d02f7843941851518089c54ffc826361cb9d6d627b5bcfec9d5269, uid : exit status 1: 2024/02/27 15:13:58 socat[45310] E connect(5, AF=2 127.0.0.1:8443, 16): Connection refused
```

kosmoz:
Closes #302

## 📑 Description
This PR changes the opener to choose the first ready pod matching the services label selector, rather than simply the first one. If no pod is ready, it fails.

## ✅ Checks

### DIFF
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -198,7 +198,12 @@ func (o *opener) pod(ctx context.Context, service *corev1.Service) (*corev1.Pod,
 	if len(pods.Items) < 1 {
 		return nil, fmt.Errorf("no pod found for service %v", service.Name)
 	}
-	return &pods.Items[0], nil
+	for _, pod := range pods.Items {
+		if isPodReady(pod) {
+			return &pod, nil
+		}
+	}
+	return nil, fmt.Errorf("no ready pod found for service %v", service.Name)
 }
 
 func portMapping(service *corev1.Service, pod *corev1.Pod, entrypoint v1alpha1.PackageEntrypoint) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/open/open.go
```diff
@@ -198,7 +198,12 @@ func (o *opener) pod(ctx context.Context, service *corev1.Service) (*corev1.Pod,
 	if len(pods.Items) < 1 {
 		return nil, fmt.Errorf("no pod found for service %v", service.Name)
 	}
-	return &pods.Items[0], nil
+	for _, pod := range pods.Items {
+		if isPodReady(pod) {
```
christophenne, line 202: This could be simplified with `IsStatusConditionTrue(pod.Status.Conditions, corev1.ConditionTrue)` from `k8s.io/apimachinery/pkg/api/meta/conditions.go`

christophenne (changes requested):

pkg/open/open.go
kosmoz: Unfortunately not. I was surprised too, but `corev1.PodCondition` is distinct from `metav1.Condition`

christophenne (approved):

PR #332, opened 2024-03-05
feat(package-operator): add handling of packages with helm and manifests

### LINKED ISSUE #328: Support additional manifests
Using cert-manager.io as an alternative to the `keptn-certifcate-operator` requires the manual creation of certificates as described in the [official keptn documentation](https://keptn.sh/stable/docs/installation/configuration/cert-manager/).

Allow the possibility to inject local manifests which can be stored next to the package manifest file in addition to manifests hosted on GitHub.

In the case of `keptn` these manifests need to be created before the `HelmRelease` creation, else the installation will not succeed.

Instead of using `localManifests` we could have used `manifestUrls` and host them in our package repository and make sure manifests and helm releases can get installed together.

**Additional context**
This feature blocks: https://github.com/glasskube/packages/pull/32 and is required for the `keptn` installation guide.

kosmoz:
Closes #328 

## 📑 Description

This PR changes the way the package controller operators pretty significantly: Previously, we updated the package status intermittently in some cases, but deferred updates to the end, risking missing data in case of an error, in others. Both approaches are not ideal were used inconsistently. 

With this change, the package status is only updated once per reconciliation. The update is performed as the last step, whether an error has occurred or not. This means that we no longer lose track of owned resources that were created before an error occurred. 

HelmReleases and HelmRepositories created by the FluxHelmAdapter are now also tracked as owned resources.

It is now possible to support both helm and manifests in a package. This allows package authors to include additional resources not provided by the helm chart. 

## ✅ Checks

### DIFF
--- a/cmd/package-operator/main.go
+++ b/cmd/package-operator/main.go
@@ -35,6 +35,7 @@ import (
 	packagesv1alpha1 "github.com/glasskube/glasskube/api/v1alpha1"
 	"github.com/glasskube/glasskube/internal/controller"
 	"github.com/glasskube/glasskube/internal/manifest/helm/flux"
+	"github.com/glasskube/glasskube/internal/manifest/plain"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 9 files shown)


christophenne (commented):
LGTM, but there is an issue when cleaning up after cert-manager installation: 
"cleanup failed: could not prune resource: Object 'Kind' is missing in 'unstructured object has no kind'; could not prune resource: Object 'Kind' is missing in 'unstructured object has no kind'"

kosmoz:
> LGTM, but there is an issue when cleaning up after cert-manager installation: "cleanup failed: could not prune resource: Object 'Kind' is missing in 'unstructured object has no kind'; could not prune resource: Object 'Kind' is missing in 'unstructured object has no kind'"

Thanks, I will investigate :detective:

kosmoz:
I updated the code in `ToOwnedResourceRef`. It is now a little less ergonomic but it can fail with an error, so this bug should not happen again.

christophenne (approved):

PR #336, opened 2024-03-06
feat(cli): add --no-await in uninstall command

### LINKED ISSUE #133: Add CLI-Flag to not await install/uninstall commands

hanshal101:
Fixes #133 

## 📑 Description
Add CLI-Flag to not await install/uninstall commands

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/uninstall.go
+++ b/cmd/glasskube/cmd/uninstall.go
@@ -14,6 +14,7 @@ import (
 
 var uninstallCmdOptions = struct {
 	ForceUninstall bool
+	NoWait         bool
 }{}
 
 var uninstallCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/glasskube/cmd/uninstall.go
```diff
@@ -39,15 +40,26 @@ var uninstallCmd = &cobra.Command{
...
+			if uninstallCmdOptions.NoWait {
+				if err := uninstaller.Uninstall(cmd.Context(), &pkg); err != nil {
+					fmt.Fprintf(os.Stderr, "An error occurred during uninstallation:\n\n%v\n", err)
+					os.Exit(1)
+				}
+				fmt.Fprintln(os.Stderr, "Uninstallation started in background")
+			} else {
+				if err := uninstall.NewUninstaller(client).
+					WithStatusWriter(statuswriter.Spinner()).
+					UninstallBlocking(cmd.Context(), &pkg); err != nil {
```
kosmoz, line 56:
```suggestion
				if err := uninstaller.UninstallBlocking(cmd.Context(), &pkg); err != nil {
```

kosmoz (changes requested):
Thanks @hanshal101! This looks pretty good. 

I just have one suggestion, could you please take a look?

kosmoz:
By the way, I noticed that you don't show up in our contributors list yet. This is probably because of the email address you use for your git commits. If you want to show up in this list, you can change the email address like this (replace the email address and name with your own):
* for new commits using `git config user.email "email@address.com"`
* for this existing commit using `git commit --amend --author="Author Name <email@address.com>" --no-edit` and then `git push --force-with-lease`

hanshal101:
Ohh I didn't noticed this. Will do this. should I commit once more with my email and the above instructions above?

PR #343, opened 2024-03-08
feat: mandatory package version (#341)

### LINKED ISSUE #341: Mandatory Package Version
In #311 we found that the auto-updating process handling is not the correct way to approach this. Instead we want a separate component (operator) doing such updates by patching the Package version, instead of the package to have an empty version. See also #338. 

Because of this change, the Package Version should be mandatory in the schema, and consequently all places where the Package Version is being used (operator, CLI, UI) must be refactored to the new logic. A package that should be auto-updated should get a label "glasskube/auto-updated". A package with such a label will then be picked up by the new controller (#338). 

Documentation must be updated too.

None

**Additional context**
See #311, #338

christophenne:
Closes #341

## ✅ Checks


api/v1alpha1/package_types.go
```diff
@@ -27,8 +27,9 @@ type PackageInfoTemplate struct {
 	// Name of the package to install
 	// +kubebuilder:validation:Required
 	Name string `json:"name"`
-	// Version of the package to install (optional)
-	Version string `json:"version,omitempty"`
+	// Version of the package to install
+	// +kubebuilder:validation:Required
```
kosmoz, line 31: I think this can be omitted, since controller-gen infers this from the omission of the "omitempty" tag. (Please check if this is true)

pkg/install/install.go
```diff
@@ -28,29 +28,42 @@ func (obj *installer) WithStatusWriter(sw statuswriter.StatusWriter) *installer
...
+	ctx context.Context,
+	packageName,
+	version string,
+	autoUpdates bool,
+) (*v1alpha1.Package, error) {
 	obj.status.SetStatus(fmt.Sprintf("Installing %v...", packageName))
 	pkg := client.NewPackage(packageName, version)
+	if autoUpdates {
+		pkg.SetLabels(map[string]string{
+			"glasskube-auto-updated": "true",
```
kosmoz, line 64: It is common to prefix labels with a domain to prevent collisions. I think something like `packages.glasskube.dev/auto-update: "true"` would be more idiomatic. 

Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
Examples: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

kosmoz (changes requested):
Some minor comments. Overall good cleanup :broom:

pkg/install/install.go
christophenne: Ok, I thought I'd skip the prefix because it makes the label "private", but it probably doesn't matter.

api/v1alpha1/package_types.go
christophenne: Yes you're right.

kosmoz (approved):

PR #346, opened 2024-03-11
feat(cli): show entrypoints in glasskube describe command

### LINKED ISSUE #297: Show entrypoints in `glasskube describe`
The user doesn't see which entrypoints an installed package has.

Show the entrypoints of the installed manifest and the hint that these can be opened with `glasskube open`.
In the describe command, the installed manifest is already being used to display some information. The goal of this ticket is to also display a list of entrypoints, which are stored in the manifest. 

**Additional context**

hanshal101:
Closes #297 
Fixes #297 

## 📑 Description
Added entrypoints in the `glasskube desscribe` command

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/describe.go
+++ b/cmd/glasskube/cmd/describe.go
@@ -46,14 +46,30 @@ var describeCmd = &cobra.Command{
 			fmt.Printf(" * %v: %v\n", ref.Label, ref.Url)
 		}
 
-		fmt.Println("\nEntrypoints:")
+		fmt.Printf("\n%v\n", bold("Entrypoints:"))
 		if len(*entrypoints) == 0 {
-			fmt.Println("No Entry Points")
+			fmt.Fprintln(os.Stderr, " * No Entry Points")
 		} else {
 			for _, i := range *entrypoints {
-				fmt.Fprintf(os.Stderr, "Name: %s | ServiceName: %s | Port: %v | LocalPort: %v | Scheme: %s\n", i.Name, i.ServiceName, i.Port, i.LocalPort, i.Scheme)
+				fmt.Fprintf(os.Stderr, " * ")
+				if i.Name != "" {
+					fmt.Fprintf(os.Stderr, "Name: %s, ", i.Name)
+				}
+				if i.ServiceName != "" {
+					fmt.Fprintf(os.Stderr, "ServiceName: %s, ", i.ServiceName)
+				}
+				if i.Port != 0 {
+					fmt.Fprintf(os.Stderr, "Port: %v, ", i.Port)
+				}
+				if i.LocalPort != 0 {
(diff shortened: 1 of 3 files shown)


pkg/describe/describe.go
```diff
@@ -14,26 +14,26 @@ import (
 func DescribePackage(
 	ctx context.Context,
 	pkgName string,
-) (*v1alpha1.Package, *client.PackageStatus, *v1alpha1.PackageManifest, error) {
+) (*v1alpha1.Package, *client.PackageStatus, *v1alpha1.PackageManifest, *[]v1alpha1.PackageEntrypoint, error) {
```
christophenne, line 17: You don't need to return the array additionally, it is already contained in the returned `PackageManifest`. So there should be no changes necessary to this function.

cmd/glasskube/cmd/describe.go
```diff
@@ -45,7 +45,33 @@ var describeCmd = &cobra.Command{
 		for _, ref := range manifest.References {
 			fmt.Printf(" * %v: %v\n", ref.Label, ref.Url)
 		}
+
+		fmt.Printf("\n%v\n", bold("Entrypoints:"))
+		if len(*entrypoints) == 0 {
```
christophenne, line 50: instead of `entrypoints` you can use `manifest.Entrypoints`

cmd/glasskube/cmd/describe.go
```diff
@@ -45,7 +45,33 @@ var describeCmd = &cobra.Command{
...
+					fmt.Fprintf(os.Stderr, "ServiceName: %s, ", i.ServiceName)
+				}
+				if i.Port != 0 {
+					fmt.Fprintf(os.Stderr, "Port: %v, ", i.Port)
+				}
+				if i.LocalPort != 0 {
+					fmt.Fprintf(os.Stderr, "LocalPort: %v, ", i.LocalPort)
+				}
+				if i.Scheme != "" {
+					fmt.Fprintf(os.Stderr, "Scheme: %s\n", i.Scheme)
```
christophenne, line 68: if `Scheme == ""`, the output will not contain the linebreak at the end. And also the colon would be the last sign of an entrypoint. 

I would suggest to use the `strings.Join` function here. Inside the loop you can hold an array `messageParts []string`, and for each property you can do `messageParts = append(messageParts, fmt.Sprintf("<property-name>: %s", <property-value>))` and at the end of the loop you call the join function on these message parts with the "," as delimiter.

christophenne (changes requested):
Hi, thanks for your PR. I have some comments on the implementation, so please have a look at it. 

When you are done, please also squash your commits and force-push on your branch – That way the PR and later on, our commit history, will only contain one commit.

kosmoz (approved):

christophenne (approved):

PR #348, opened 2024-03-11
feat(package-operator): add validation with webhook

### LINKED ISSUE #116: Dependent packages should not be uninstallable
A package that is a dependency of another package should not be uninstallable. This should be implemented via webhook validation. 

Note that checking owner references only will not be sufficient, because a package could have existed before the dependent package has been installed.

### LINKED ISSUE #322: Validate Package Updates via Webhook
It is not allowed to set a version on a `Package`, that is not compatible with versions that dependent packages require. This should be implemented as webhook.

kosmoz:
Closes #322  
Closes #116

## 📑 Description

This PR adds a validating admission webhook for the operator.

This webhook uses the existing DependencyManager.Validate function for validation.

For package deletion, it verifies that the package either doesn't have any owning packages, or that all owning packages have a deletion timestamp (this is to prevent a deadlock).

## ✅ Checks

## ℹ Additional Information

It is possible to run the operator on the host machine and use it for the validation webhook using an ExternalName service (only works on minikube). To run the operator including the webhook configuration, run `make install webhook cert run`.
- `install` deploys the CRDs
- `webhook` deploys the webhook config with a patch to target the host machine
- `cert` creates a self signed TLS certificate on the host machine and patches the webhook config with the CA bundle
- `run` executes the operator

When deployed, the TLS certificate is created/updated by a Job and CronJob running once per month in the glasskube-system namespace.

This PR is the continuation of #342

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,20 +6,24 @@ The goal of this document is to define some guidelines to streamline our contrib
 ## Getting started
 
 ### Where should I start?
+
 - If you are new to the project, please check out the [good first issue](https://github.com/glasskube/glasskube/labels/good%20first%20issue) label.
 - If you are looking for something to work on, check out our [open issues](hhttps://github.com/glasskube/glasskube/issues).
 - If you have an idea for a new feature, please open an issue, and we can discuss it.
 - We are also happy to help you find something to work on. Just reach out to us.
 
 ### Getting in touch with the community
+
 - Join our [Glasskube Discord Channel](https://discord.gg/SxH6KUCGH7)
 - Introduce yourself on the intros channel or open an issue to let us know that you are interested in contributing
 
 ### Discuss issues
+
 - Before you start working on something, propose and discuss your solution on the issue
 - If you are unsure about something, ask the community
 
 ### How do I contribute?
+
 - Fork the repository and clone it locally
 - Create a new branch and follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) guidelines for work undertaken
 - Assign yourself to the issue, if you are working on it (if you are not a member of the organization, please leave a comment on the issue)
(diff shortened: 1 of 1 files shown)


internal/webhook/package_validator.go
```diff
@@ -0,0 +1,129 @@
...
+// to at least one other package. However, we do have to allow packages to be deleted if all their dependant
+// packages are already being deleted, because the webhook is also called if a package is garbage-collected and
+// this would lead to a dead lock otherwise:
+// The dependant package is not deleted because of the foreground deletion finalizer and the dependency is not
+// deleted because the dependant is not deleted.
+func (p *PackageValidatingWebhook) ValidateDelete(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
+	log := ctrl.LoggerFrom(ctx)
+	if pkg, ok := obj.(*v1alpha1.Package); ok {
+		log.Info("validate delete", "name", pkg.Name)
+		if dependants, err := p.OwnersOfType(&v1alpha1.Package{}, pkg); err != nil {
```
christophenne, line 81: I think this is missing the case in which there is a package dependent on `pkg` without having an owner reference on it.

christophenne (changes requested):
Looking good – I tested it and it works as expected in most cases. As discussed I created #364 and found another issue at deletion (see comment).

internal/repo/client/fake/fakeclient.go
```diff
@@ -0,0 +1,95 @@
...
+		f.Packages[name] = map[string]v1alpha1.PackageManifest{}
+	}
+	f.Packages[name][version] = manifest
+}
+
+func (f *FakeClient) Clear() {
+	f.Packages = map[string]map[string]v1alpha1.PackageManifest{}
+}
+
+var _ client.RepoClient = &FakeClient{}
```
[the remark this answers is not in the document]
kosmoz: This tells the compiler that this type implements the RepoClient interface (which is not possible any other way :upside_down_face:)

internal/webhook/package_validator.go
kosmoz: This is a good point. As discussed, we will implement this in a separate PR, as a similar functionality is also needed in other places.

PR #360, opened 2024-03-13
feat: prune package dependencies (#318)

### LINKED ISSUE #318: Prune unused dependencies after package update
Given `P -> D`, but an update of `P` removes the dependency on `D`, `D` should be removed if it is not needed by any other package and if it has not been installed manually (and has not been "made" manually, see #315 ).

TBD

christophenne:
Closes #318

## ✅ Checks


internal/controller/package_controller.go
```diff
@@ -544,3 +557,58 @@ func (r *PackageReconcilationContext) pruneOwnedPackageInfos(ctx context.Context
...
+	ownedPackagesCopy := r.pkg.Status.OwnedPackages[:]
+OuterLoop:
+	for _, ref := range r.pkg.Status.OwnedPackages {
+		for _, newRef := range r.currentOwnedPackages {
+			if ownerutils.RefersToSameResource(ref, newRef) {
+				continue OuterLoop
+			}
+		}
+
+		r.setShouldUpdate(ownerutils.Remove(&ownedPackagesCopy, ref))
```
kosmoz, line 573: I think it would be more resilient to only remove the reference if we know that either:
- it doesn't exist anymore (get returns 404 error)
- it doesn't have an owner reference to this pkg anyways
- the update removing the owner reference has succeeded

In the edge case of the update succeeding but the deletion after removing the last owner reference failing, I think it would be OK for now that the package remains as an orphan.

kosmoz (changes requested):

kosmoz (approved):

PR #375, opened 2024-03-14
feat(package-operator): prevent invalid updates and deletions (#364)

### LINKED ISSUE #364: Prevent invalid updates
The dependency manager checks only dependencies "downwards" (i.e. dependencies of a package), but not "upwards" (i.e. packages that are dependent on a package). This means that currently it's possible to update a package to a version that is invalid for a dependant package, leading to a dependency conflict. 

Also check parent packages during validation. 

**Additional context**

christophenne:
Closes #364

## ✅ Checks

### DIFF
--- a/internal/dependency/manager.go
+++ b/internal/dependency/manager.go
@@ -7,6 +7,9 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/glasskube/glasskube/internal/controller/owners"
+	"github.com/glasskube/glasskube/internal/controller/owners/utils"
+
 	"github.com/glasskube/glasskube/internal/dependency/adapter"
 
 	"github.com/Masterminds/semver/v3"
@@ -18,6 +21,7 @@ import (
 type DependendcyManager struct {
 	clientAdapter adapter.ClientAdapter
 	repoAdapter   adapter.RepoAdapter
+	*owners.OwnerManager
 }
 
 type ValidationResultStatus string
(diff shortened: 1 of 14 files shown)


internal/dependency/manager.go
```diff
@@ -191,3 +196,103 @@ func (dm *DependendcyManager) createRequirement(dependency v1alpha1.Dependency)
...
+}
+
+func (dm *DependendcyManager) getDependency(ctx context.Context, pkg *v1alpha1.Package, dependency string) (*dependentTuple, error) {
+	pkgInfo, err := dm.clientAdapter.GetPackageInfo(ctx, pkg.Status.OwnedPackageInfos[0].Name)
+	if err != nil {
+		return nil, err
+	}
+	if pkgInfo.Status.Manifest == nil {
+		return nil, err
+	}
```
kosmoz, line 288: In this case, `nil, nil` would be returned. If this is desired, I think it would be better to return `nil, nil` explicitly.

internal/webhook/package_validator.go
```diff
@@ -105,9 +101,11 @@ func (p *PackageValidatingWebhook) validateCreateOrUpdate(ctx context.Context, p
 
 	if result, err := p.Validate(ctx, pkg, &manifest); err != nil {
 		return err
-	} else if len(result.Conflicts) > 0 {
+	} else if parentConflicts, err := p.DependendcyManager.IsUpdateAllowed(ctx, pkg, pkg.Spec.PackageInfo.Version); err != nil {
```
kosmoz, line 104: Minor, but `p.IsUpdateAllowed(...)` should work too.

kosmoz (changes requested):
Some minor comments, but it works really well!

Issue #291, opened 2024-02-26
Persistent Repo Client Cache


christophenne:
**Is your feature request related to a problem? Please describe.**
We request some resources of the package repository quite often, these could be cached. In one case this has been implemented already, see here: https://github.com/glasskube/glasskube/pull/286#discussion_r1502544043 . But this served just as a temporary solution, and we want a solution that is easier to re-use for other repo caching purposes too. 

**Describe the solution you'd like**
* Research caching solutions and libraries
* Implement caching for queries to the package repo for CLI commands and package operator.

**Describe alternatives you've considered**

**Additional context**

kosmoz:
We now have a in-memory cache for the repo client (#381), which especially helps the serve command and package-operator to perform much less requests.

To improve this even further, we should investigate implementing a shared on-disk cache.

PR #381, opened 2024-03-15
feat(ui): add syncing update notification via websocket

### LINKED ISSUE #355: Update not working when global update notification not present
**Describe the bug**
When updating from the package overview page, while the update notification is not present yet, the update does not do anything. This scenario shouldn't happen in the first place, but it happens because the update notification is not properly synced with the update buttons per package. 

**To Reproduce**
1. Open package overview and make sure the "global" update notification is not present. 
2. Install a package, not in latest version but in an older one.
3. When the package is installed, the update button appears
4. Clicking the update button opens the modal correctly, but the update button there does not trigger the update.

**Expected behavior**
Update should work, but also the global update notification should be properly in sync with the update button per package. 

 - Type: Kubernetes
 - Version: 
 - Provider: minikube

 - Package Operator version: latest
 - CLI version: latest

**Additional context**

Browser console says: "htmx:targetError". This means that the backend request is not even sent by htmx, because the target (the div of the update notification) is not present. This is why nothing happens.

### LINKED ISSUE #357: Sync global update notification correctly
Global update notification is only evaluated when the package overview page is rendered, and it might be (incorrectly) hidden dynamically by the update dialog. However, this is wrong and has many edge cases. 

Instead, like the update buttons of each package, the global update notification should appear/disappear at every package status change. For that, we need to hold all installed packages on the server and evaluate the state at every status change. 

**Additional context**
This will fix #355

kosmoz:
Closes #355
Closes #357 

## 📑 Description
This PR a cached variant of the package client, which a regular package client can be transparently swapped for. This makes calling updater.Prepare() much less expensive, so it can be called for every package update/page load. 

I also made the informer update func more resilient by also broadcasting the state if the installedManifest can not be retrieved. This fixes the error of sometimes incorrect button states being shown, e.g. right after uninstalling a package.

## ✅ Checks


internal/web/components/pkg_overview_btn/controller.go
```diff
@@ -28,23 +28,25 @@ func getButtonId(pkgName string) string {
...
-		Swap:            fmt.Sprintf("outerHTML:#%s", buttonId),
-		PackageName:     pkg.Name,
-		Status:          status,
-		Manifest:        manifest,
+		ButtonId:    buttonId,
+		Swap:        fmt.Sprintf("outerHTML:#%s", buttonId),
+		PackageName: pkg.Name,
+		Status:      status,
+		Manifest:    manifest,
+		// TODO: Use semver check
```
christophenne, line 36: Same in the `pkg_detail_btns` – I think instead of comparing the semver versions, we should use the `s.updateAvailable` logic in these places too. At broadcasting time we now calculate it already anyway, so for the other two components this could be passed as parameter in the `Render` functions.

When rendering the package overview page, we also have it available already (only globally and not per package, but that's almost there I would say). It would only need to be called when rendering the package detail page. 

But it's fine to do this in a separate PR/issue. 

Also related, since we are now calling the updater more often (and with the above suggestion even more often), we should cache at least the package index calls to the repo too (`defaultClient.FetchPackageRepoIndex`). It doesn't have to be that sophisticated, I think something simple like it already is done in `defaultClient.GetLatestVersion` would be enough for now. We have a ticket for a general approach #291.

christophenne (commented):
Great solution and works perfectly fine for me!

See only the comment regarding some follow-up work.

internal/web/components/pkg_overview_btn/controller.go
kosmoz: Good idea. I did both (also added repo caching)

christophenne (changes requested):
Found one issue with the package detail page, then it's good to go.

internal/web/server.go
```diff
@@ -304,41 +319,37 @@ func (s *server) packages(w http.ResponseWriter, r *http.Request) {
...
-		"CurrentContext": s.rawConfig.CurrentContext,
-		"Package":        pkg,
-		"Status":         status,
-		"Manifest":       manifest,
-		"LatestVersion":  latestVersion,
+		"CurrentContext":  s.rawConfig.CurrentContext,
+		"Package":         pkg,
+		"Status":          status,
+		"Manifest":        manifest,
+		"UpdateAvailable": s.isUpdateAvailable(r.Context(), pkgName),
```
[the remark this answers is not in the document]
kosmoz: Good catch. I fixed this error and changed `DescribePackage` to return `nil` for the package and status if the package is not installed.

I also added a TODO comment to remove the `latestVersion` return parameter in `DescribePackage` since I think it's a little misplaced there.

christophenne (approved):

PR #386, opened 2024-03-18
feat(ui): handle dependencies (#114)

### LINKED ISSUE #114: Show package dependencies in UI
The UI should provide a way for the user to see which dependencies a package includes.

christophenne:
Ref #114

## 📑 Description
In this PR we add showing package dependencies on the package detail page, showing the list of installed packages (including dependencies) at the installation modal, and showing a list of dependency conflicts in case installation is not possible. 

## ✅ Checks

### DIFF
--- a/internal/web/templates.go
+++ b/internal/web/templates.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path"
 
+	"github.com/glasskube/glasskube/api/v1alpha1"
+
 	"github.com/glasskube/glasskube/internal/repo"
 
 	"github.com/glasskube/glasskube/internal/web/components/pkg_detail_btns"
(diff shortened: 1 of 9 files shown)


internal/web/templates.go
```diff
@@ -32,8 +34,14 @@ func init() {
 	templateFuncs := template.FuncMap{
 		"ForPkgOverviewBtn": pkg_overview_btn.ForPkgOverviewBtn,
 		"ForPkgDetailBtns":  pkg_detail_btns.ForPkgDetailBtns,
-		"PackageManifestUrl": func(pkgName string) string {
-			if url, err := repo.GetPackageManifestURL("", pkgName, ""); err != nil {
+		"PackageManifestUrl": func(pkgName string, pkg *v1alpha1.Package, latestVersion string) string {
```
christophenne, line 37: We have linked to the wrong manifest URL, this is the fix.

kosmoz (approved):
Looks good. Just one small suggestion: Since this PR only adds handling of dependencies during installation (not for updates), I think the commit message should reflect that.

PR #389, opened 2024-03-18
docs: add instructions for installation via nix

jakuzure:
## 📑 Description

Added instructions on how to install Glasskube via nix, this works for Linux and MacOS (and maybe Windows), so I added it to the Readme directly as opposed to the installation methods web page.

This PR is still marked as a draft, because the Nixpkgs PR has not been merged yet, please leave a 👍 to increase the chances of it being merged quicker: https://github.com/NixOS/nixpkgs/pull/285513

## ✅ Checks

### DIFF
--- a/website/src/partials/_install.mdx
+++ b/website/src/partials/_install.mdx
@@ -51,6 +51,14 @@ import TabItem from '@theme/TabItem';
     latest [Release](https://github.com/glasskube/glasskube/releases/latest) and unpack it using Windows Explorer.
     {/* x-release-please-end */}
   </TabItem>
+
+  <TabItem value="nixos" label="NixOS/Nixpgks">
+    You can either use the package temporarily in a <b>nix-shell</b>:
+    ```
+    nix-shell -p glasskube
+    ```
+    Or install it globally by adding `pkgs.glasskube` to your `environment.systemPackages`.
+  </TabItem>
 </Tabs>
 
 <br/>


kosmoz:
Hi @jakuzure! 

I took a look at your PR over at https://github.com/NixOS/nixpkgs/pull/285513. I'm not familiar with Nix packages but it looks like you build Glasskube from source. This is possible, of course, but some things won't work out of the box. To make everything work, you need to run `make web` or `npm i; npm run build` before building the go binary. This also requires a working node.js installation.

Alternatively, you can also download release binaries from our GitHub releases but I don't know if this approach would be acceptable for the Nix folks.

jakuzure:
Hi @kosmoz,

I originally tested it with v0.0.1 and the UI was looking fine. Thanks for your input, the PR above was merged in the meantime and I created another that copies the npm build artifacts over before doing the go build here: https://github.com/NixOS/nixpkgs/pull/298338

Using already built binaries is usually the last resort in Nix (or done for software where no source is available), building from source is strongly preferred

kosmoz (changes requested):
Hi @jakuzure! I see that your second PR in the nixpkg repository also got merged. We would be happy to add nix to our documentation, however, we also aim to keep the README file as concise as possible, so I think the installation guide in our documentation (`website/docs/02_getting-started/01_install.mdx`) would be a more appropriate place for this. 

Could you please update your PR and move the instructions to the install docs? :slightly_smiling_face:

jakuzure:
Hi @kosmoz, does this work for you?

pmig (commented):
Thanks for the adaption. It might be easier for users if we  create a separat `<Tab>` with the label `NixOS/Nixpgks` and value `nixos`

jakuzure:
> Thanks for the adaption. It might be easier for users if we create a separat `<Tab>` with the label `NixOS/Nixpgks` and value `nixos`

Good idea, I added a separate tab

kosmoz (approved):
Looks good to me :+1:

PR #395, opened 2024-03-19
fix(cli): update check fails in dev environment

### LINKED ISSUE #396: Flag --skip-update-check fails in dev build
**Describe the bug**
When the flag --skip-update-check is used in cli, it should check for updates and print the update message but the flag seems to fail in dev builds.

**To Reproduce**
Use the --skip-update-check flag followed by a command
`glasskube <command> --skip-update-check=false`

 - Package Operator version: 0.0.4
 - CLI version: dev

christophenne:
Closes #396

## ✅ Checks

## ℹ Additional Information
Thanks @Baalekshan for bringing this up!

### DIFF
--- a/internal/cliutils/checkforupdate.go
+++ b/internal/cliutils/checkforupdate.go
@@ -14,6 +14,8 @@ import (
 func CheckForUpdate() (*string, error) {
 	if releaseInfo, err := releaseinfo.FetchLatestRelease(); err != nil {
 		return nil, err
+	} else if config.IsDevBuild() {
+		return &releaseInfo.Version, nil
 	} else if version, err := semver.NewVersion(config.Version); err != nil {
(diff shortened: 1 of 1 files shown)


internal/cliutils/checkforupdate.go
```diff
@@ -27,7 +29,9 @@ func CheckForUpdate() (*string, error) {
 
 func UpdateFetch() {
 	newerVersion, err := CheckForUpdate()
-	if err == nil && newerVersion != nil {
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "There was an error checking for updates: %v\n", err)
```
kosmoz, line 33: I think the check should continue to fail silently, as it is not really related to what the user is doing

PR #401, opened 2024-03-19
feat(cli, ui): show dependencies when updating

### LINKED ISSUE #113: Show package dependencies in CLI
The CLI should provide a way for the user to see which dependencies a package includes.

### LINKED ISSUE #114: Show package dependencies in UI

christophenne:
Closes #113 
Closes #114

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -65,6 +65,9 @@ func printTransaction(tx update.UpdateTransaction) {
 				item.Package.Name, item.Package.Spec.PackageInfo.Version)
 		}
 	}
+	for _, req := range tx.Requirements {
+		fmt.Fprintf(w, "%v:\t-\t-> %v\n", req.Name, req.Version)
+	}
 	_ = w.Flush()
 }
 
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -5,17 +5,24 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/glasskube/glasskube/api/v1alpha1"
+	"github.com/glasskube/glasskube/internal/controller/owners"
+	"github.com/glasskube/glasskube/internal/dependency"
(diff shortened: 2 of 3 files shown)


cmd/glasskube/cmd/update.go
```diff
@@ -60,6 +60,9 @@ func printTransaction(tx update.UpdateTransaction) {
 		if item.UpdateRequired() {
 			fmt.Fprintf(w, "%v:\t%v\t-> %v\n",
 				item.Package.Name, item.Package.Spec.PackageInfo.Version, item.Version)
+			for _, req := range tx.Requirements {
+				fmt.Fprintf(w, "%v:\t-\t-> %v\n", req.Name, req.Version)
+			}
```
kosmoz, line 65: I think this should be done after the for loop over tx.Items, since the dependencies are collected on the transaction, not per item. If there are multiple items, the dependencies would be listed for each item.

pkg/update/update.go
```diff
@@ -5,17 +5,25 @@ import (
...
 	"github.com/glasskube/glasskube/pkg/condition"
 	"github.com/glasskube/glasskube/pkg/statuswriter"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
 )
 
 type UpdateTransaction struct {
 	Items []updateTransactionItem
+	// TODO: use ConflictItems and Requirements in update command
```
kosmoz, line 24: This TODO is no longer accurate, since the requirements are already considered :slightly_smiling_face:

kosmoz (changes requested):

cmd/glasskube/cmd/update.go
christophenne: I had one job

kosmoz (approved):

PR #404, opened 2024-03-20
fix: consider semver metadata when comparing installed vs latest version (#397)

### LINKED ISSUE #397: Build numbers are not considered in version comparison
**Describe the bug**
When creating a new latest version of a package in the package repo, which only differs to its predecessor in the build number (i.e. previously latest version: 2.7.0+1, new latest version: 2.7.0.+2), the package is not marked as updatable. 

**To Reproduce**
0. use a local package repo
1. install kubernetes-dashboard in 2.7.0+1 for simplicity via CLI
2. in a local package repo, copy version 2.7.0+1 of kubernetes-dashboard, with version number 2.7.0+2
3. make 2.7.0+2 the latest version in the repo index and package index
4. run `glasskube update`: kubernetes-dashboard will not be in the list of updatable packages.

**Expected behavior**
`glasskube update` should probably show kubernetes-dashboard as updatable, even though this is not perfectly following the semver specification (see additional context below).

 - Type: Kubernetes
 - Version: 1.29
 - Provider: minikube

 - Package Operator version: dev
 - CLI version: dev

**Additional context**
The underlying comparison of versions is done with the semver package, and according to semver spec, this comparison is correct: the build number (or what we interpret as build number) could be any string. See https://github.com/helm/helm/issues/6710#issuecomment-558711082

kosmoz:
Closes #397 

## 📑 Description
With this PR, the build metadata field of the semver is also considered when we check if a version can be upgraded. For this I defined the following rules:

* installed and latest have same metadata (empty or not) -> false
* installed has metadata and latest has no metadata -> false
* installed has no metadata and latest has metadata -> true
* installed and latest have a number as metadata -> installed < latest
* installed and latest have metadata but either is not a number -> installed != latest

Whether we want to allow metadata values that are not integers is of course up for debate, but at least the behavior in such a case is now defined.

## ✅ Checks

### DIFF
--- a/internal/semver/compare_test.go
+++ b/internal/semver/compare_test.go
@@ -1,0 +1,64 @@
+package semver
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("IsUpgradable", func() {
+	// testCases is a map of installed version to latest version to expected result
+	testCases := map[string]map[string]bool{
+		"v1.0.0": {
+			"v1.0.0":        false,
+			"v1.0.0+1":      true,
+			"v1.0.0+a":      true,
+			"not a version": true,
+		},
+		"v1.0.0+1": {
+			"v1.0.0":        false,
+			"v1.0.0+1":      false,
+			"v1.0.0+2":      true,
+			"v1.0.0+a":      true,
+			"not a version": true,
+		},
+		"v1.0.0+a": {
+			"v1.0.0":        false,
+			"v1.0.0+1":      true,
+			"v1.0.0+a":      false,
+			"v1.0.0+b":      true,
+			"not a version": true,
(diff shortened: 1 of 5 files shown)


internal/semver/compare_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("IsUpgradable", func() {
+	// testCases is a map of installed version to latest version to expected result
+	testCases := map[string]map[string]bool{
```
christophenne, line 12: Since I am doing something similar, I stumbled across [Table Specs](https://onsi.github.io/ginkgo/#table-specs) for such cases – but it's only syntactic sugar and not necessary to change here.

christophenne (approved):

PR #410, opened 2024-03-20
fix(ui): make the whole card on package list clickable

kosmoz:
## 📑 Description

I noticed that only the texts and buttons on the package list are actual links. With this PR, the entire card surface becomes clickable.

## ✅ Checks


christophenne (commented):
When clicking "Install" now the dialog opens and in the background the page changes to the package detail page. Not sure if this is wanted, also debatable whether correct or not. 

Same for the Open button, it does the opening operation, but also redirects you to the detail page.

kosmoz:
> When clicking "Install" now the dialog opens and in the background the page changes to the package detail page. Not sure if this is wanted, also debatable whether correct or not.
> 
> Same for the Open button, it does the opening operation, but also redirects you to the detail page.

interesting, I'll check again

christophenne (approved):

PR #413, opened 2024-03-21
feat(ui): trigger refreshing package detail page (#382)

### LINKED ISSUE #382: Re-Render Package Detail Page after install, uninstall and update
A user might install a different from the latest version of some package from the package detail page. When they entered this page, it was rendered based on the manifest of the latest version. After installation, update or uninstallation, the page is outdated (e.g. manifest link is wrong, dep because it shows a different version. 

The detail part of the page should be reloaded in some smart htmx way. 

**Additional context**

christophenne:
Closes #382

## 📑 Description
We make use of the `Hx-Trigger` response header at the `install`, `update` and `uninstall` endpoints and use this trigger on the package detail page to reload its content.

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -51,6 +51,8 @@ import (
 //go:embed templates
 var embededFs embed.FS
 
+const TriggerRefreshPackageDetail = "gk:refresh-package-detail"
+
 type ServerOptions struct {
 	Host       string
 	Port       int32
@@ -215,7 +217,9 @@ func (s *server) install(w http.ResponseWriter, r *http.Request) {
 		Install(r.Context(), pkgName, selectedVersion, strings.ToLower(enableAutoUpdateVal) == "on")
 	if err != nil {
 		s.respondAlertAndLog(w, err, "An error occurred installing "+pkgName)
+		return
 	}
+	addHxTrigger(w, TriggerRefreshPackageDetail)
 }
 
 func (s *server) installModal(w http.ResponseWriter, r *http.Request) {
@@ -310,6 +314,7 @@ func (s *server) update(w http.ResponseWriter, r *http.Request) {
 		s.respondAlertAndLog(w, err, "An error occurred updating"+pkgName)
 		return
 	}
+	addHxTrigger(w, TriggerRefreshPackageDetail)
 }
 
 func (s *server) uninstall(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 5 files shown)


kosmoz (commented):
Unfortunately, this doesn't work for me. After clicking "Install", the UI only shows a blank page. I think this might be because the page is refreshed before the PackageInfo has been created and reconciled by the controller. Since we use asynchronous installation in the UI (which I would keep this way), there is no easy way to determine when the PackageInfo becomes available.

Additionally, I found that the package list page is broken if a package doesn't have a PackageInfo, but that's probably not a result of this PR, so I will create a ticket. It just panics at this line: 
https://github.com/glasskube/glasskube/blob/a86fc04017ef0066ad289abac110cfcfcfedd827/pkg/list/list.go#L124

EDIT: I created #419 to track this

christophenne:
> Unfortunately, this doesn't work for me. After clicking "Install", the UI only shows a blank page. I think this might be because the page is refreshed before the PackageInfo has been created and reconciled by the controller. Since we use asynchronous installation in the UI (which I would keep this way), there is no easy way to determine when the PackageInfo becomes available.
> 
> Additionally, I found that the package list page is broken if a package doesn't have a PackageInfo, but that's probably not a result of this PR, so I will create a ticket. It just panics at this line:
> 
> https://github.com/glasskube/glasskube/blob/a86fc04017ef0066ad289abac110cfcfcfedd827/pkg/list/list.go#L124
> 
> EDIT: I created #419 to track this

Ok I'll look into it again, I did not run into this issue. 
Agree about keeping it asynchronous, the idea was only to show the detail page in the appropriate version after the installation or update. It's not a big deal now, but I think it will become more apparent once there are configuration forms etc. on the page, which depend on the version. But maybe this is not the right way to do this after all.

kosmoz:
> Ok I'll look into it again, I did not run into this issue.
> Agree about keeping it asynchronous, the idea was only to show the detail page in the appropriate version after the installation or update. It's not a big deal now, but I think it will become more apparent once there are configuration forms etc. on the page, which depend on the version. But maybe this is not the right way to do this after all.

I've been thinking about this too, lately. I'm actually not sure if keeping install information and "UI stuff" in the same place is the right approach in the first place. Maybe, moving the logo, descriptions etc. to the versions.yaml file (or a different file altogether) would be better?

christophenne:
> > Ok I'll look into it again, I did not run into this issue.
> > Agree about keeping it asynchronous, the idea was only to show the detail page in the appropriate version after the installation or update. It's not a big deal now, but I think it will become more apparent once there are configuration forms etc. on the page, which depend on the version. But maybe this is not the right way to do this after all.
> 
> I've been thinking about this too, lately. I'm actually not sure if keeping install information and "UI stuff" in the same place is the right approach in the first place. Maybe, moving the logo, descriptions etc. to the versions.yaml file (or a different file altogether) would be better?

"in the same place" as in the package repository you mean? yes i could imagine splitting the metadata from the rest, but i would still see the metadata in a separate file in each version instead of the versions.yaml. 

Let's discuss it on Monday, I think there might be some follow-up issues too like "version dropdown on package detail page" (which could also extend in "update to this version" for a specific version on the UI). I'll leave this PR unchanged for now until we decide how to proceed.

christophenne:
Decision: Quickfix

kosmoz (approved):
Works great :+1:

PR #418, opened 2024-03-22
feat(cli): update to specific package version cli

### LINKED ISSUE #412: Update to specific package version in CLI
A user can only update an outdated version to the latest version. However when they are on a version newer than latest (e.g. some prerelease of the next major version) they cannot update anymore. Also when dependency requirements make it impossible to update to the latest version from an older version, it might still make sense to update to a version below latest. 

The `update` command should provide a `-v` version flag like the `install` command. Invalid updates to a smaller version than the one installed must be prohibited.
This version flag is only allowed in combination with updating a single package, as it would make no sense in the "update all" scenario.

Right now there are no plans to offer this on the UI explicitly. 

**Additional context**
Related #344 (the missing glue will probably be "calculating" the maximum possible version to update to)

hanshal101:
Closes #412
Fixes #412 

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -47,7 +47,6 @@ var updateCmd = &cobra.Command{
 					fmt.Fprintf(os.Stderr, "⛔ Update cancelled. No changes were made.\n")
 					os.Exit(0)
 				}
-
 				if err := updater.Apply(ctx, tx); err != nil {
 					fmt.Fprintf(os.Stderr, "❌ update failed: %v\n", err)
 					os.Exit(1)
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -17,6 +17,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var updateCmdOptions struct {
+	Version string
+}
+
 var updateCmd = &cobra.Command{
 	Use:               "update [packages...]",
 	Short:             "Update some or all packages in your cluster",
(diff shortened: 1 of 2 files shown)


cmd/glasskube/cmd/update.go
```diff
@@ -29,24 +33,45 @@ var updateCmd = &cobra.Command{
...
+				if !cliutils.YesNoPrompt("Do you want to apply these updates?", false) {
+					fmt.Fprintf(os.Stderr, "⛔ Update cancelled. No changes were made.\n")
+					os.Exit(0)
+				}
+
+				if err := updater.Apply(ctx, tx); err != nil {
+					fmt.Fprintf(os.Stderr, "❌ update failed: %v\n", err)
+					os.Exit(1)
+				}
+			}
```
christophenne, line 74: this whole part does not need to be duplicated, it's the same for both decision branches.

pkg/update/update.go
```diff
@@ -70,6 +70,42 @@ func (c *updater) WithStatusWriter(writer statuswriter.StatusWriter) *updater {
 	return c
 }
 
+func (c *updater) VersionUpdate(ctx context.Context, pkgName string, pkgVersion string) (*UpdateTransaction, error) {
```
christophenne, line 73: please call it something similar to `PrepareForVersion`, so that one sees it's doing something similar as the `Prepare` function

christophenne (changes requested):
thanks for the PR, please have a look at the comments.

pkg/update/update.go
hanshal101: for `PrepareForVersion` it is showing the lint error of line exceeded

christophenne: but then you can reformat the line to be in multiple lines, for example

```
func (c *updater) PrepareForVersion(
  ctx context.Context,
...
) (...) {
```

hanshal101:
I hope this commit solves all the previous issue.

christophenne:
> I hope this commit solves all the previous issue.

It's a good step, but I think you removed the Yes/No Prompt and the call to `updater.Apply(ctx, tx)` by accident, so now the user is not asked to confirm + the updates are not applied.

christophenne (changes requested):
Please have a look at the remaining comment. Then I will test it and if everything works, we are good to go here :)

PR #442, opened 2024-03-27
chore(website): update author avatar urls

pmig:
## 📑 Description
streamlining avatar urls, updating louis linkedin link

## ✅ Checks


kosmoz (approved):
Nit: I think the commit message should be of type "chore" rather than "docs", since the docs do not actually change.

PR #445, opened 2024-03-27
feat(client, ui): introduce package info cache (#444)

### LINKED ISSUE #444: Cached PackageInfo Client
The package client is already cached, the package info one not yet. This came up in #415, where we noticed that when having at bigger number of packages (at least 4) results in a lot of calls to get package infos from the API server. The go client throttles the requests to the server with its [default settings](https://pkg.go.dev/k8s.io/client-go/rest#pkg-constants). 

Apply the same caching mechanism to the package info client to reduce the number of requests.

In the specific use case mentioned above, we could also tweak the client side throttling settings to allow more requests.

**Additional context**

christophenne:
Closes #444

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -85,17 +85,19 @@ func NewServer(options ServerOptions) *server {
 type server struct {
 	ServerOptions
 	configLoader
-	listener           net.Listener
-	restConfig         *rest.Config
-	rawConfig          *api.Config
-	pkgClient          client.PackageV1Alpha1Client
-	wsHub              *WsHub
-	informerStore      cache.Store
-	informerCtrl       cache.Controller
-	forwarders         map[string]*open.OpenResult
-	dependencyMgr      *dependency.DependendcyManager
-	updateMutex        sync.Mutex
-	updateTransactions map[int]update.UpdateTransaction
+	listener              net.Listener
+	restConfig            *rest.Config
+	rawConfig             *api.Config
+	pkgClient             client.PackageV1Alpha1Client
+	wsHub                 *WsHub
+	packageStore          cache.Store
+	packageController     cache.Controller
+	packageInfoStore      cache.Store
+	packageInfoController cache.Controller
+	forwarders            map[string]*open.OpenResult
+	dependencyMgr         *dependency.DependendcyManager
+	updateMutex           sync.Mutex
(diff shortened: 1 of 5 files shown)


kosmoz (changes requested):
I'm getting some errors after startup:
```
Error fetching manifest for package argo-cd:  "argo-cd--v2.10.4--1" not found
Error fetching manifest for package cyclops:  "cyclops--v0.2.0--1" not found
```
But everything works and the error does not occur again after a page reload.

Might be related to the informer controller not being started explicitly for PackageInfos?

christophenne:
> I'm getting some errors after startup:
> 
> ```
> glasskube UI is available at http://localhost:8580
> Error fetching manifest for package argo-cd:  "argo-cd--v2.10.4--1" not found
> Error fetching manifest for package cyclops:  "cyclops--v0.2.0--1" not found
> ```
> 
> But everything works and the error does not occur again after a page reload.
> 
> Might be related to the informer controller not being started explicitly for PackageInfos?

Thanks for the hint that the controller has not been started, fixed that.

The other errors seem to be unrelated, at least to this change here. But I can reproduce these errors quite deterministically on the main branch too. I don't know since when we have that. From a first analysis it seems to me that this happens always when we install a package or update a package. At installation the package info or the manifest might not exist yet, so we fail to fetch them in the informer update func, or in the broadcasting function when checking for updates respectively. 
They don't have an "ultimate" consequence on the UI because there is always a following update where the manifest can be fetched eventually.

kosmoz:
> The other errors seem to be unrelated, at least to this change here. But I can reproduce these errors quite deterministically on the main branch too. I don't know since when we have that. From a first analysis it seems to me that this happens always when we install a package or update a package. At installation the package info or the manifest might not exist yet, so we fail to fetch them in the informer update func, or in the broadcasting function when checking for updates respectively.
> They don't have an "ultimate" consequence on the UI because there is always a following update where the manifest can be fetched eventually.
> That being said, we should still handle this somehow.

I think they were definitely related because they are gone now :slightly_smiling_face: But I know that this was not the only cause of this error. I think it's actually fine not to handle this error explicitly, but the logging might be a little noisy.

kosmoz (approved):

PR #449, opened 2024-03-28
feat(ui): display latest version in package overview

### LINKED ISSUE #330: Display installed version of a package in the UI
It is not possible to determine which package is currently installed via the UI.

- Similar to the `list`  command the installed version of a package should be displayed on the overview and detail page.
- On the detail page also display the latestVersion.
- If an update is available display the update path e.g. `v0.0.1` -> `v0.0.2`

Also always display the latest version.

hanshal101:
Closes #330
Fixes #330

## 📑 Description
This shows the LatestVersion and the Current version (if installed) on the package overview page
Images :
- When a Package is installed 
image
- When a Package is not installed 
image

## ✅ Checks

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,60 +1,79 @@
-# Contributing guide for Glasskube
+# Contributing guide for Glasskube 🧊
 
 Welcome, and thank you for deciding to invest some of your time in contributing to the Glasskube project!
 The goal of this document is to define some guidelines to streamline our contribution workflow.
 
-## Getting started
+## Before you get started ✋
+---
+There are many types of issues you can take on when contributing to the Glasskube project. We try our best to provide a wide array of open issues that vary in levels of complexity. From beginners to seasoned developers, everyone should be able to find something to work on.
 
-### Where should I start?
+### Let's find the perfect open issue for you!
 
 - If you are new to the project, please check out the [good first issue](https://github.com/glasskube/glasskube/labels/good%20first%20issue) label.
-- If you are looking for something to work on, check out our [open issues](hhttps://github.com/glasskube/glasskube/issues).
+- If you are ready to make a big impact on the project, check out the [current milestone](https://github.com/glasskube/glasskube/milestones) that is being worked on and filter the issues by `"help-wanted"`, these issues are the ones that will make it into the next official release. 
+- If you are looking for something specific to work on, check out our [open issues](https://github.com/glasskube/glasskube/issues?q=is%3Aissue+is%3Aopen+no%3Aassignee+-label%3Aneeds-triage) and filter against the available tags such as `component: cli`, `component: ui` `component: repo`, `documentation`.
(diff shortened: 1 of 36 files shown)


kosmoz (changes requested):
Looks good!

I think it would make sense to only show the "latest version" if it would make sense to upgrade to it. For this, one would have to register `semver.IsUpgradable` as a `templateFuncs` in `internal/web/templates.go` so it can be called from a template. Then it should be possible to use it in a template like this:
```html
{{ if IsUpgradable .Package.Status.Version .LatestVersion }}
<!-- show latest version here -->
{{ end }}
```

Do you think you could incorporate this change @hanshal101?

hanshal101:
I can try this, can you give me some more information on this. And what's the actual purpose it solves.

hanshal101:
Also I forgot the path which sends all the data to the template. Could you please guide me to the file. (Data : Manifests, Status, Package etc. In the package-detail page)

kosmoz:
> Also I can't find the path which sends all the data to the template. Could you please guide me to the file. (Data : Manifests, Status, Package etc. In the package-detail page)

I think what you're looking for is the `packageDetail` function in `internal/web/server.go`.

> what's the actual purpose it solves.

I think, as a user I don't really care what the latest version is if I'm already using it and especially if I am using a version that is _newer_ than the latest version (the latest version is not necessarily the highest available, it's the one that was tested and reviewed).

> I can try this, can you give me some more information on this. 

If you have any more specific questions I can try to answer them! :slightly_smiling_face:

hanshal101:
Well from your explaination I understood :
- Adding a `IsUpgrade` in the `templatesFunc`
Looking something like this:
```
"IsUpgradeble": semver.IsUpgradable,
```
- After performing this I added the htmx code as : 
```
<span>
            {{ if IsUpgradable .Package.Status.Version .LatestVersion }}
              <i> Latest Version: {{ .LatestVersion }} </i>
            {{ end }}
            <br />
            {{ if ne .Status nil }}
              <i> Current Version: {{ .Package.Status.Version }} </i>
            {{ end }}
</span>
```

Is this something we are looking for?
Here I have a small doubt. If I apply this code then we wont be able to see the `latest version` in the package which is not installed.

kosmoz:
That looks like it might work. I think `IsUpgradable` returns true if one of the versions is empty so we should be fine here.

hanshal101:
What about showing `LatestVersion` on packages which are not installed?

kosmoz:
> What about showing `LatestVersion` on packages which are not installed?

In this case, the latest version should be shown. But I think your code already does this.

hanshal101:
Hey here I made a few mistakes before in the commits but fixed it in the next commit. But still the checks have not passed as it shows the error of the past commit too. Is there any way to fix it.

kosmoz:
`git fetch origin && git rebase origin/main`
… then test your changes, then …
`git push --force-with-lease`

PR #450, opened 2024-03-29
feat(api): add types for value configurations

kosmoz:
## 📑 Description
Adds types for the specification described in #446

## ✅ Checks
- [x] I have updated the documentation as required (#446)

## ℹ Additional Information
Uses some CEL validation features ([reference](https://kubernetes.io/docs/reference/using-api/cel/))


config/crd/bases/packages.glasskube.dev_packages.yaml
```diff
@@ -64,6 +64,59 @@ spec:
...
+                      - message: ValueReference must have exactly one of configMapRef,
+                          secretRef, packageRef
+                        rule: '[has(self.configMapRef), has(self.secretRef), has(self.packageRef)].exists_one(x,
+                          x)'
+                  type: object
+                  x-kubernetes-validations:
+                  - message: ValueConfiguration must have exactly one of value, valueFrom,
+                      packageRef
+                    rule: '[has(self.value), has(self.valueFrom)].exists_one(x, x)'
+                type: object
```
christophenne, line 119: these validations seem to be problematic when applying them in a cluster with `make install`:

```
The CustomResourceDefinition "packages.packages.glasskube.dev" is invalid: 
* spec.validation.openAPIV3Schema.properties[spec].properties[values].additionalProperties.properties[valueFrom].x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.8x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
* spec.validation.openAPIV3Schema.properties[spec].properties[values].additionalProperties.x-kubernetes-validations[0].rule: Forbidden: estimated rule cost exceeds budget by factor of 2.3x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
```

kosmoz: Good catch! I found a different solution that works for these specific cases.

PR #452, opened 2024-03-30
feat(ui): display latest version in package overview

### LINKED ISSUE #330: Display installed version of a package in the UI

hanshal101:
Closes #330
Fixes #330

## 📑 Description
This shows the LatestVersion and the Current version (if installed) on the package overview page
Images :
- When a Package is installed 
image
- When a Package is not installed 
image

## ✅ Checks

### DIFF
--- a/internal/web/templates/pages/package.html
+++ b/internal/web/templates/pages/package.html
@@ -83,11 +83,10 @@
               <i>Current Version: {{ .Package.Status.Version }}</i>
             </div>
           {{ end }}
-          </div>
+        </div>
         {{ if ne .Manifest.Dependencies nil }}
           <div class="mt-3">
             <h2 class="text-reset">Package Dependencies</h2>
-
             <ul>
               {{ range .Manifest.Dependencies }}
                 <li>
--- a/internal/web/templates/pages/package.html
+++ b/internal/web/templates/pages/package.html
@@ -73,17 +73,17 @@
               {{ .Manifest.LongDescription }}
             {{ end }}
           </span>
-          <br />
-          <span>
-            {{ if IsUpgradable .Package.Status.Version .LatestVersion }}
-              <i> Latest Version: {{ .LatestVersion }} </i>
-            {{ end }}
-            <br />
-            {{ if ne .Status nil }}
-              <i> Current Version: {{ .Package.Status.Version }} </i>
-            {{ end }}
-          </span>
-        </div>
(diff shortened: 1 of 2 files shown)


hanshal101:
@kosmoz Here, I think there is an error when the package is not installed and we hit the `package detail` page The error is being seen in the terminal

internal/web/templates/pages/package.html
```diff
@@ -73,8 +73,17 @@ <h2 class="text-reset">Package Details</h2>
               {{ .Manifest.LongDescription }}
             {{ end }}
           </span>
+          <br />
+          <span>
+            {{ if IsUpgradable .Package.Status.Version .LatestVersion }}
```
kosmoz, line 78: As you already discovered, there is a nil pointer error here. You can fix it by changing the expression to `if or (eq .Package nil) (IsUpgradable .Package.Status.Version .LatestVersion)`

internal/web/templates/pages/package.html
```diff
@@ -73,8 +73,17 @@ <h2 class="text-reset">Package Details</h2>
...
+              <i> Latest Version: {{ .LatestVersion }} </i>
+            {{ end }}
+            <br />
+            {{ if ne .Status nil }}
+              <i> Current Version: {{ .Package.Status.Version }} </i>
+            {{ end }}
+          </span>
```
kosmoz, line 85: I think it would be preferable to use block elements instead of forced line breaks. Something like this:

```suggestion
          {{ if IsUpgradable .Package.Status.Version .LatestVersion }}
            <div>
              <i>Latest Version: {{ .LatestVersion }}</i>
            </div>
          {{ end }}
          {{ if ne .Status nil }}
            <div>
              <i>Current Version: {{ .Package.Status.Version }}</i>
            </div>
          {{ end }}
```

internal/web/templates/pages/package.html
hanshal101: Yup this solves the issue.

internal/web/templates/pages/package.html
hanshal101: Ok will do the necessary changes.

kosmoz (approved):

PR #460, opened 2024-04-03
feat(package-operator): add patching resources, helm values with package values

### LINKED ISSUE #118: Support package configuration in operator and repo
The package repo has to provide a way to make parts of a package configurable. The fields and of which type they can be, needs to be speced first.

kosmoz:
Closes #118 

## 📑 Description
TODO

## ✅ Checks


christophenne (approved):
Good job! I was able to test both helm and manifest in combination with what I currently have on the UI. Scaling up a second replica already worked for both cases. 

I only have one question (see comment) – in case that is not a concern I think it can be merged.

christophenne (approved):

PR #461, opened 2024-04-03
feat(cli): update to specific package version cli

### LINKED ISSUE #412: Update to specific package version in CLI

hanshal101:
Closes #412 
Fixes #412 

## 📑 Description
This code adds the `-v` flag in the update section. Using this user can update to any version of the package they desire.

## ✅ Checks

## ℹ Additional Information
I have tried replicating the `Prepare` function but for a single package.😊

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -17273,9 +17273,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.4.3",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.4.3.tgz",
-      "integrity": "sha512-KrPd3PKaCLr78MalgiwJnA25Nm8HAmdwN3mYUYZgG/wizIo9EainNVQI9/yDavtVFRN2h3k8uf3GLHuhDMgEHg==",
+      "version": "5.4.4",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.4.4.tgz",
+      "integrity": "sha512-dGE2Vv8cpVvw28v8HCPqyb08EzbBURxDpuhJvTrusShUfGnhHBafDsLdS1EhhxyL6BJQE+2cT3dDPAv+MQ6oLw==",
       "bin": {
         "tsc": "bin/tsc",
         "tsserver": "bin/tsserver"


cmd/glasskube/cmd/update.go
```diff
@@ -47,6 +60,7 @@ var updateCmd = &cobra.Command{
 				fmt.Fprintf(os.Stderr, "❌ update failed: %v\n", err)
 				os.Exit(1)
 			}
+			printTransaction(*tx)
```
christophenne, line 63: `printTransaction(*tx)` prints the transaction to the terminal, so that the user can decide whether they want to apply these updates or not. Therefore it should be called before the user is asked to confirm (`YesNoPrompt`).

pkg/update/update.go
```diff
@@ -65,6 +65,60 @@ func (c *updater) WithStatusWriter(writer statuswriter.StatusWriter) *updater {
...
 
+func (c *updater) PrepareForVersion(
+	ctx context.Context,
+	pkgName string,
+	pkgVersion string) (*UpdateTransaction, error) {
+	c.status.Start()
+	defer c.status.Stop()
+	c.status.SetStatus("Collecting installed package")
+
+	var packagesToUpdate []v1alpha1.Package
```
christophenne, line 76: We don't need this in `PrepareForVersion`. There is always only one package to update. So you don't need to create this slice, and you don't need to iterate over it in line 91. It's just one package always.

pkg/update/update.go
```diff
@@ -65,6 +65,60 @@ func (c *updater) WithStatusWriter(writer statuswriter.StatusWriter) *updater {
...
+	if err := repo.FetchPackageRepoIndex("", &index); err != nil {
+		return nil, fmt.Errorf("failed to fetch index: %v", err)
+	}
+
+	requirementsSet := make(map[dependency.Requirement]struct{})
+	var tx UpdateTransaction
+	for _, pkg := range packagesToUpdate {
+		for _, indexItem := range index.Packages {
+			if indexItem.Name == pkg.Name {
+				item := updateTransactionItem{Package: pkg, Version: pkgVersion}
```
christophenne, line 94: Please have a look at the similar function `Prepare`: At this point, it is important to check whether the requested version is newer than the currently installed version. Downgrade should not be possible.

cmd/glasskube/cmd/update.go
```diff
@@ -29,13 +33,22 @@ var updateCmd = &cobra.Command{
...
+		var err error
 
-		printTransaction(*tx)
+		if len(args) == 1 && updateCmdOptions.Version != "" {
+			tx, err = updater.PrepareForVersion(ctx, args[0], updateCmdOptions.Version)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error in updating the package version : %v", err)
+				os.Exit(1)
+			}
+		} else {
```
christophenne, line 45: In case the user enters more than one package and the `-v` flag, I think it would be nice to show them an error message: "Version flag not supported for multiple packages" or something like that. And after that, `os.Exit`.

christophenne (changes requested):
Hi, thanks for your PR. Please have a look at the comments. 

When you are done, please execute the following test scenarios to make sure it works as intended: 

* Install cert-manager in version 1.13.3+1. Then use the update command with the version flag to update to 1.14.2+1 – this should work.
* When the previous update has succeeded, try to update the package down to 1.13.+1 with the version flag. This should not work, i.e. no update should be made here. 
* Try to update multiple packages with the version flag, e.g. `glasskube update cert-manager keptn ingress-nginx -v 1.0.0`. This should not work, i.e. there should be an error message. 

Also, when all of this works, please make sure to squash your commits. Thanks a lot!

cmd/glasskube/cmd/update.go
hanshal101: image
Do we need something like this?

christophenne: Fine by me, but I would remove the " : <nil>" part

pkg/update/update.go
hanshal101: Can you please guide me on this. According to me we should use the `semver.IsVersionUpgradable` function.

christophenne: That's exactly the way. Use this function, and if it returns true, you can continue, otherwise you can return because the update is not possible.

hanshal101: Well I tried using this code:
```
parsedLatest, err := s.NewVersion(indexItem.LatestVersion)
			if err != nil {
				fmt.Fprintf(os.Stderr, "Error in creating the parsed version\n")
			}
			parsedDesired, err := s.NewVersion(pkgVersion)
			if err != nil {
				fmt.Fprintf(os.Stderr, "Error in creating the parsed version\n")
			}
			if !semver.IsVersionUpgradable(parsedLatest, parsedDesired) {
				fmt.Fprintf(os.Stderr, "Can't update downgraded version\n")
				return nil, nil
			}
```

But here when I am updating from the upper-version to the latest version it is stopping that

christophenne: What is upper version and what is latest version in your scenario?

hanshal101: Upper version: argo-cd v2.10.5+1
Latest Version: argo-cd v2.10.4+1

christophenne: Ok, and which one is currently installed?

hanshal101: Upper Version

christophenne: Then it works exactly as intended. 

You have argo-cd v2.10.5+1 installed, and argo-cd v2.10.4+1 is of course older than your currently installed version. This means you cannot update to v2.10.4+1.

hanshal101: Ya but v2.10.4+1 is the latest version right? and users can update to the latest version.

christophenne: Sorry for the confusion about the word "latest", but that's just where we are right now. 

Latest version in our context means, it is the version which is "marked as latest" in the package repository. See for example: https://github.com/glasskube/packages/blob/main/packages/argo-cd/versions.yaml . Here, 2.10.4+1 is marked as latest, but clearly 2.10.5+1 is newer according to semantic versioning. 

For your feature, it does not matter at all what is marked as latest version. The only thing that matters is, you cannot "update" to a version which is smaller than the currently installed version.

hanshal101: Ohk understood will do this test and then commit the code:
https://github.com/glasskube/glasskube/pull/461#pullrequestreview-1978859558

hanshal101:
Well, I have tested the above steps. Everything works fine. There were few issues in the code.
Changes:
- Here:
```
parsedLatest, err := s.NewVersion(pkg.Status.Version)
			if err != nil {
				return nil, fmt.Errorf("error parsing latest version: %v", err)
			}
			parsedDesired, err := s.NewVersion(pkgVersion)
			if err != nil {
				return nil, fmt.Errorf("error parsing desired version: %v", err)
			}
			if !semver.IsVersionUpgradable(parsedLatest, parsedDesired) {
				return nil, fmt.Errorf("can't update to downgraded version or equal version")
			}
```

We need to add `pkg.Status.Version` instead of `indexItem.LatestVersion` as it should show the latest-version instead of current-version of the package.

christophenne (changes requested):
Some small final steps, then we are done. 

Also, please fix your commit message, the current one is wrong.

hanshal101:
I hope this will solve the issue.

christophenne:
@hanshal101 do you need help with this?

hanshal101:
So are there any conflicts still there or is it ready to merge?

christophenne:
There are no conflicts but it is not ready to merge because of the commits. Please have a look here: https://github.com/glasskube/glasskube/pull/461/commits – There is a second commit from Jake, which somehow contains parts your changes. There should be only one commit (with your commit message), that contains all of your changes.

christophenne:
continued in #489

PR #483, opened 2024-04-05
feat(cli): should not show latest version if installed version is newer

### LINKED ISSUE #441: `glasskube ls` should not show latest available version if installed version is newer than latest
Output of `glasskube ls` when having installed keptn in a later version than latest:
```
 go run cmd/glasskube/main.go ls
NAME                  STATUS         VERSION                              AUTO-UPDATE
argo-cd               Not installed                                         
cert-manager          Ready          v1.14.2+1                              
cyclops               Not installed                                         
ingress-nginx         Not installed                                         
keptn                 Ready          v2.0.0-rc.1+1 (v0.10.0+1 available)    
kubernetes-dashboard  Not installed                                         
```

In the above scenario, "(v0.10.0+1 available)" should not be shown because the version marked as latest is older than the version installed. 

**Additional context**

hanshal101:
Closes #441
Fixes #441

## 📑 Description
`glasskube ls` should not show latest available version if installed version is newer than latest version
This is how the terminal looks like now.
Notice here the argo-cd package is installed on the pre-release version.
```
PS H:\glasskube> .\main.exe ls   
NAME                  STATUS         VERSION                                  AUTO-UPDATE
argo-cd               Ready          v2.10.5+1
cert-manager          Ready          v1.14.2+1
cyclops               Not installed
ingress-nginx         Not installed
keptn                 Ready          v2.0.0-rc.1+1 (v2.0.0-rc.2+1 available)
kubernetes-dashboard  Ready          v2.7.0+1 (v2.7.0+2 available)
```
## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -130,7 +130,7 @@ func versionString(pkg list.PackageWithStatus) string {
 			if statusVersion != specVersion {
 				versionAddons = append(versionAddons, fmt.Sprintf("%v desired", specVersion))
 			}
-			if repoVersion != "" && semver.IsUpgradable(statusVersion, repoVersion) && statusVersion != repoVersion {
+			if repoVersion != "" && semver.IsUpgradable(statusVersion, repoVersion) {
 				versionAddons = append(versionAddons, fmt.Sprintf("%v available", repoVersion))
 			}
 			if len(versionAddons) > 0 {
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/glasskube/glasskube/api/v1alpha1"
 	"github.com/glasskube/glasskube/internal/cliutils"
+	"github.com/glasskube/glasskube/internal/semver"
 	"github.com/glasskube/glasskube/pkg/client"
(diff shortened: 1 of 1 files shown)


cmd/glasskube/cmd/list.go
```diff
@@ -129,7 +130,7 @@ func versionString(pkg list.PackageWithStatus) string {
 			if statusVersion != specVersion {
 				versionAddons = append(versionAddons, fmt.Sprintf("%v desired", specVersion))
 			}
-			if repoVersion != "" && statusVersion != repoVersion {
+			if repoVersion != "" && !semver.IsUpgradable(repoVersion, statusVersion) && statusVersion != repoVersion {
```
kosmoz, line 133: I see two issues here:
1. I think `semver.IsUpgradable(statusVersion, repoVersion)` would make more sense than `!semver.IsUpgradable(repoVersion, statusVersion)`, since we want to know whether the actual version could be upgraded to the one from the repo, not the other way around.
2. `statusVersion != repoVersion` should be omitted, since this would make the condition true, even if IsUpgradable evalutes to false

kosmoz (changes requested):

cmd/glasskube/cmd/list.go
hanshal101: ohk

kosmoz (approved):

PR #494, opened 2024-04-10
feat(ui): show warning if operator and client versions differ (#352)

### LINKED ISSUE #352: Show and resolve version mismatch between client and operator on the UI
We have the check in place once #205 is done, but there is no way to see it for the UI user. 

If glasskube is bootstrapped, but the operator is in an older version, the user should get redirected to a page where they see the hint that their operator is outdated (similar to what they see on the CLI commands). Like on the regular bootstrap page, there should be a button to upgrade. 

Needs discussion about what to do with "custom" installations (i.e. slim-bootstrapped environments). 

**Additional context**
Related #205

christophenne:
Closes #352

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/version.go
+++ b/cmd/glasskube/cmd/version.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/glasskube/glasskube/internal/clientutils"
+
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/config"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


kosmoz (commented):
The code changes look good, but I have some comments on the visuals:
1. I think it would look better to replace the emoji with a bootstrap icon.
2. I'm not sure if an end user should be expected to know what a package-operator is. Maybe using "glasskube cluster version" would be more understandable.
3. I think the message could be more actionable, e.g. 
   > Please update the Glasskube cluster components to 0.2.0 by running `glasskube bootstrap`.

kosmoz (approved):

PR #500, opened 2024-04-11
feat(cli): version completion when updating to specific version

### LINKED ISSUE #462: CLI version completion when updating to specific version
Once #412 is done, the `glasskube update` command is able to update a package to a specific version. However, it does not yet support shell completion yet like the install command does. 

When the user enters `glasskube update <package-name> -v` and then presses the tab key, a list of possible versions should appear, from which the user can choose. This is similar to what the install command already does when the `-v` flag is given, however its not exactly the same: the update command completion should only show those versions of the package, which are newer then the currently installed version. 

**Additional context**
Follow up of #412

As an entrypoint, you can take a look at the `install` command, specifically the function `completeAvailablePackageVersions`. It will have to be adapted to only show versions that are newer than the currently installed one (meaning that the package also needs to be fetched first).

To be fair, its a trickier "good first issue". As always, you can contact us any time in case of questions or problems.

hanshal101:
Closes #462 
Fixes #462 

## 📑 Description

Here I have added the auto completion feature in the update version command, This will update only to those versions which are greater than the current version.

## ✅ Checks

## ℹ Additional Information

I tried this once in my local setup. But failed to try it again as there were some issues with my system.

### DIFF
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"slices"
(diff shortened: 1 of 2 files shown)


christophenne:
@hanshal101 could you please make sure, that
* the build checks pass
* there is exactly one commit in the PR describing your changes (squash your commits)

PR #530, opened 2024-04-19
fix(ui): restructure server initializations

christophenne:
## 📑 Description
This PR fixes a performance issue which occurs when at least 4 packages are installed in a cluster + the user switches back and forth between package detail pages very quickly. The cause of this was that the used package client was not actually the cached one, because of the incorrect initialization. 

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -107,6 +107,7 @@ type server struct {
 	updateMutex           sync.Mutex
 	updateTransactions    map[int]update.UpdateTransaction
 	valueResolver         *manifestvalues.Resolver
+	isBootstrapped        bool
 }
 
 func (s *server) RestConfig() *rest.Config {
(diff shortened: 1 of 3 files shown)


internal/web/server.go
```diff
@@ -682,30 +690,33 @@ func (server *server) initKubeConfig() ServerConfigError {
...
-	server.pkgClient = client
+	server.pkgClient = client // be aware that server.pkgClient is overridden with the cached client once bootstrap check succeeded
+	return nil
+}
+
+func (server *server) initWhenBootstrapped(ctx context.Context) {
+	k8sclient, err := kubernetes.NewForConfig(server.restConfig)
+	if err != nil {
+		panic(err)
+	}
```
kosmoz, line 705: This is the same as NewForConfigOrDie

kosmoz (commented):
Code looks good, but I get a strange error when installing a package:
image

It looks like the installation failed, although it actually didn't. I think this is because we try to get the package info before the informer cache has been updated. :confused:

christophenne:
> Code looks good, but I get a strange error when installing a package: image
> 
> It looks like the installation failed, although it actually didn't. I think this is because we try to get the package info before the informer cache has been updated. 😕

Good point, didn't see it because it doesn't always appear, but often enough. Tried to fix it.

kosmoz (approved):

PR #484, opened 2024-04-07
Improve error handling

Horiodino:
Closes # 

## ✅ Checks
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required

## ℹ Additional Information

in `internal/controller/package_controller.go` need to do changes , does the changes make any sence ? or i should modify it

Issue #60, opened 2024-01-18
Improve error handling for package installation


christophenne:
Currently, when the installation of a package fails, the install command fails as well, printing an error message to the user.

* Example 1: `glasskube install nginx-ingress` could fail because of some network error or unavailable package repository. The user would see the error message, unaware that in the background the controllers will try the installation over and over again until the package repository is back up. This is an example of an error, where it could be beneficial for the user, when the command blocks until the error is gone. 
* Example 2: `glasskube install non-existent-package` will always fail, and therefore retrying will never work. The user should see the error message again.

Things that have already been mentioned:
* In any error case, the command line tool could delete the recently created `Package` custom resource to break the reconciliation loop. 
* Alternatively we could listen to the "type" of error and whether it is resolvable by retrying. 
* Once we have an `wait` or `no-wait` option, and the user chooses not to wait, they won't get the end result anyway.

We should discuss the best possible error handling strategy here, based on different kind of errors and use cases.

Horiodino:
Hii , i recently saw this project and i was wondering if no one is working on it , can i !
and any other context like where to made changes  or any other references.

kosmoz:
Hi @Horiodino, are you still interested?

Some pointers for anyone who wants to pick this one up:

The CLI needs to determine from the package status whether a Failed condition is recoverable or not. As mentioned in the issue description, the way to do this would probably be via the reason field of the condition. We already have a `Reason` type for this at `pkg/condition/condition.go` where we could add a `func (r Reason) Recoverable() bool` that the client could then use after converting the `condition.Reason` (which is a string) to a `Reason`. We probably also need to add more fine-grained reasons for this to really be useful.

kosmoz:
Alright, let us know if you have any questions!

kosmoz:
Hi @Horiodino! How are you doing? Are you still working on this issue?

christophenne:
I linked https://github.com/glasskube/packages/pull/232 because maybe that kind of error is "recoverable".

PR #535, opened 2024-04-20
refactor: pkg recoverable condition #60

Horiodino:
Closes # 

## 📑 Description

https://github.com/glasskube/glasskube/pull/484

## ✅ Checks
- [✅] make all

## ℹ Additional Information

so what i was thinking , in `internal/controller/package_controller.go`    `PackageReconcilationContext` assuming its giving pkg related info , so what we can do is if we get any error we can read the `PackageReconcilationContext.isRecoverable` whcih is updated by the SetFailed , additionaly for detailed error im not sure how i can add it with `Recoverable` function so if there is any logger in glasskube maybe we can use that for detailed errors.

### DIFF
--- a/internal/controller/conditions/conditions.go
+++ b/internal/controller/conditions/conditions.go
@@ -61,7 +61,7 @@ func SetReadyAndUpdate(ctx context.Context, client client.Client, recorder recor
 	return nil
 }
 
-func SetFailed(ctx context.Context, recorder record.EventRecorder, obj client.Object, objConditions *[]metav1.Condition, reason condition.Reason, message string) bool {
+func SetFailed(ctx context.Context, recorder record.EventRecorder, obj client.Object, objConditions *[]metav1.Condition, reason condition.Reason, message string) (bool, bool) {
 	log := log.FromContext(ctx)
 	log.V(1).Info("set condition to failed: " + message)
 	recorder.Event(obj, "Warning", string(reason), message)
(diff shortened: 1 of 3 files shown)


internal/controller/conditions/conditions.go
kosmoz, line 1: I think this change should be reverted. The controller shouldn't care whether a failure is recoverable, we rely on the runtime to schedule retries with an exponential back-off.

pkg/condition/condition.go
```diff
@@ -21,3 +21,14 @@ const (
...
+
+func (r Reason) Recoverable() bool {
+	switch r {
+	case InstallationFailed:
+		return false
+	case UnsupportedFormat:
+		return false
+	default:
+		return true
+	}
```
kosmoz, line 33: I don't think defaulting to recoverable is a good idea. The goal of this feature would be having the ability to consider some errors as recoverable, which is different from the current behavior where all errors are considered non-recoverable.

kosmoz (changes requested):
Hi @Horiodino! Please see the comments below. As noted in my review (https://github.com/glasskube/glasskube/pull/484#pullrequestreview-1985789797) on your previous PR, I think the benefit of this feature would be on the client side, not the controller. In more concrete terms the place where `IsRecoverable()` should be evaluated would be in e.g. the `installer`. So, the controller shoudn't do anything different but the client gains the ability to differentiate between a failure that can be temporary and one that is terminal and stop waiting in such a case. 

Please let me know if this explanation is clear of if you have any further questions.

kosmoz:
Hi @Horiodino! Yes, I think you got that right.

Absolutely no problem and good luck with your exams! We might open this issue up for contributions in the meantime, so if it is fixed by the time you come back, feel free to pick one of our other [open issues](https://github.com/glasskube/glasskube/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22)! :slightly_smiling_face: 

I'll close this PR for now - you can simply create a new one once you feel ready.

Issue #529, opened 2024-04-18
`glasskube version` should not show an error when not bootstrapped


christophenne:
**Is your feature request related to a problem? Please describe.**
`glasskube version` shows an error when not bootstrapped yet, but it's a valid case to not have it bootstrapped yet and check the client version.

**Describe the solution you'd like**
Instead of the error it should show an information that it's not bootstrapped yet, and a hint that it can be done with the `bootstrap` command. 

**Describe alternatives you've considered**

**Additional context**

Aryakoste:
Hello @christophenne. I would like to work on this issue.

PR #538, opened 2024-04-21
glasskube version show an error when not bootstrapped

Aryakoste:
Closes https://github.com/glasskube/glasskube/issues/529

## 📑 Description

Changed the error message to show that the glasskube is not bootstrapped yet do it using glasskube bootstrap command.

## ✅ Checks


kosmoz (commented):
Hi @Aryakoste, thanks for your contribution! :rocket: 

I think it would be nice to have a message that is more consistent with the non-error output. something like this:

```sh
# normal state
$ glasskube version
glasskube: v0.2.0
package-operator: v0.2.0

# error state
$ glasskube version
glasskube: v0.2.0
package-operator: not installed
Glasskube is not yet bootstrapped. Use 'glasskube bootstrap' to get started.
```

Do you think you can make this change?

Aryakoste:
I have submitted new PR with the requested changes. Because my current branches caused some problem.

PR #543, opened 2024-04-22
feat(cli): changing the local port for glasskube open

### LINKED ISSUE #523: Changing the local port for `glasskube open`
There might be issues if the local port is already in use when using the `open` command. It should be possible to change the default if desired.

On the CLI, a different port can be specified with `--port`. If the flag is used and there are multiple entrypoints for the package and no single entrypoint is specified via CLI argument, the command should exit with an error. If the flag is not specified, the default from the manifest should be used.

Adding logic to pick a different random port (similar to what we already to for the `serve` command), but I think this would be more complicated an error prone in this situation, since the listener can not be reused. We might have to extract the port from the listener, then close it and reuse the same port for another listener… OTOH there is no reason why both features shouldn't exist.

**Additional context**
No idea how to do this on the UI :thinking:

hanshal101:
Closes #523 
Fixes #523 

## 📑 Description

This code adds a new feature of specifing the local port for any packages with one entrypoint, It shows an error when there is more than one entrypoint.

## Outputs
- When a specific port is mentioned:
```
PS H:\glasskube> go run .\cmd\glasskube\main.go open kubernetes-dashboard --port 3004
         |I| Forwarding from 127.0.0.1:3004 -> 8443
         |I| Forwarding from [::1]:3004 -> 8443
✅ kubernetes-dashboard is now reachable at https://localhost:3004
         |I| Handling connection for 3004
         |I| Handling connection for 3004
```
## ✅ Checks


kosmoz (changes requested):
Looks good! I just noticed one missing check (see below).

hanshal101:
Made the changes 👍

kosmoz (approved):

PR #557, opened 2024-04-23
fix(cli): prevent accidental bootstrapping of an older package-operat…

### LINKED ISSUE #304: Prevent accidental bootstrapping of an older package-operator version
If the latest newer version of the package operator is installed in the Kubernetes cluster the `bootstrap` command should just get executed, which will lead to a breaking change.

Add a confirmation input [y/N] and ask the user if he is sure that we wants to downgrade the package-operator and potentially break the system.

Completely disallow it.

**Additional context**
This ticket is related to: #205
It is still possible install a specific manifest via the `--url`  flag.

hanshal101:
Closes #304 
Fixes #304

## 📑 Description

This PR fixes the accidental bootstrapping of an older package-operator version.

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/glasskube/glasskube/internal/clientutils"
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/config"
+	"github.com/glasskube/glasskube/internal/semver"
 	"github.com/glasskube/glasskube/pkg/bootstrap"
 	"github.com/spf13/cobra"
 )
@@ -31,6 +33,35 @@ var bootstrapCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg, _ := cliutils.RequireConfig(config.Kubeconfig)
 		client := bootstrap.NewBootstrapClient(cfg)
+
+		installedVersion, err := clientutils.GetPackageOperatorVersion(cmd.Context())
+		if err != nil {
+			IsBootstrapped, err := bootstrap.IsBootstrapped(cmd.Context(), cfg)
+			if err != nil && !IsBootstrapped {
+				fmt.Printf("error : %v\n", err)
+				cliutils.ExitWithError()
+			}
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = ""
+		}
+		if !semver.IsUpgradable(installedVersion, desiredVersion) &&
+			installedVersion != "" &&
+			installedVersion[1:] != desiredVersion {
+			if !cliutils.YesNoPrompt(fmt.Sprintf("Glasskube is already installed in this cluster "+
+				"in the newer version %v. You are about to install version %v. This could lead "+
+				"to a broken cluster!\nAre you sure that you want to downgrade glasskube "+
+				"in this cluster?", installedVersion, desiredVersion), false) {
+				fmt.Println("Operation stopped")
+				cliutils.ExitWithError()
+			}
+
+		}
+
 		if err := client.Bootstrap(cmd.Context(), bootstrapCmdOptions.asBootstrapOptions()); err != nil {
 			fmt.Fprintf(os.Stderr, "\nAn error occurred during bootstrap:\n%v\n", err)
 			cliutils.ExitWithError()


christophenne (changes requested):
This prompt should not be shown at every bootstrap, only in a specific case, which is:
* The package operator is already installed in the cluster and has (for example) version v0.2.1.
* The client has version v0.2.0 (for example, too).
* When the user executes `glasskube bootstrap` in this scenario, the existing package operator would be downgraded from v0.2.1 to v0.2.0. The user should be asked in this case if they are sure about this. 

So your task is to compare the versions of client (`config.Version`) and the operator (see `clientutils.GetPackageOperatorVersion`). 
* If there is no package operator installed yet, everything is fine and you continue.
* If package operator version is equal or lower than the client version, it's also fine. 
* If package operator version is greater than the client version, then the prompt should be shown, because that would downgrade the operator, which is not a supported scenario. 

Make sure to only apply this check if the `url` flag is not given. If it is given, we cannot perform the check because we don't know which package operator version would get installed.

hanshal101:
Is this what we are expecting?

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,19 @@ var bootstrapCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg, _ := cliutils.RequireConfig(config.Kubeconfig)
 		client := bootstrap.NewBootstrapClient(cfg)
+		installedVersion, err := clientutils.GetPackageOperatorVersion(cmd.Context())
+		if err != nil {
+			fmt.Printf("error : %v\n", err)
```
christophenne, line 38: This is problematic, because if the package operator is not installed yet (i.e. glasskube is not bootstrapped), then the user would always run into this error and could never bootstrap. 

I suggest that in the case that `err != nil`, you additionally call `bootstrap.IsBootstrapped`: If the result is true, you can actually print the error (because the operator should be there, but isn't for some reason). If the result is false, continue with the regular bootstrap flow that starts here in line 49.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,19 @@ var bootstrapCmd = &cobra.Command{
...
+			return
+		}
+		desiredVersion := config.Version
```
christophenne, line 41: `config.Version` is only the desired version, if the given url is not set. So please check that `bootstrapCmdOptions.url` is empty before doing the version check. If the user sets a custom url, we cannot parse what version they are about to install.

christophenne (changes requested):

cmd/glasskube/cmd/bootstrap.go
hanshal101: Hey can you explain this, what I have done far now is:
```
var desiredVersion string
		if bootstrapCmdOptions.url == "" {
			desiredVersion = config.Version
		} else {

		}
```

cmd/glasskube/cmd/bootstrap.go
hanshal101: Does this look good:
```
IsBootstrapped, err := bootstrap.IsBootstrapped(cmd.Context(), cfg)
		if err != nil && !IsBootstrapped {
			fmt.Printf("error : %v\n", err)
			return
		}
```

christophenne: @hanshal101 please commit your current status to the branch so I can see the whole context.

cmd/glasskube/cmd/bootstrap.go
christophenne: @hanshal101 please commit your current status to the branch so I can see the whole context.

cmd/glasskube/cmd/bootstrap.go
christophenne: @hanshal101 I see, my idea was to do the bootstrap check only if `err != nil` after `GetPackageOperatorVersion()`.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,29 @@ var bootstrapCmd = &cobra.Command{
...
+			fmt.Printf("error : %v\n", err)
+			return
+		}
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = ""
+		}
+		if !semver.IsUpgradable(installedVersion, desiredVersion) {
```
christophenne, line 52: Did you test what happens if the url is set?

hanshal101:
Here I have installed the `glasskube releases v0.2.1`: 
image

And this is a wrong url `google.com`
image

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,30 @@ var bootstrapCmd = &cobra.Command{
...
+				return
+			}
+			return
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = bootstrapCmdOptions.url
```
christophenne, line 51: `desiredVersion = ""` was correct here

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,30 @@ var bootstrapCmd = &cobra.Command{
...
 		client := bootstrap.NewBootstrapClient(cfg)
+
+		installedVersion, err := clientutils.GetPackageOperatorVersion(cmd.Context())
+		if err != nil {
+			IsBootstrapped, err := bootstrap.IsBootstrapped(cmd.Context(), cfg)
+			if err != nil && !IsBootstrapped {
+				fmt.Printf("error : %v\n", err)
+				return
+			}
+			return
```
christophenne, line 44: Remove this `return` please. We don't quit when no error has happened.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,30 @@ var bootstrapCmd = &cobra.Command{
...
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = bootstrapCmdOptions.url
+		}
+		if !semver.IsUpgradable(installedVersion, desiredVersion) {
```
christophenne, line 53: in this `if` please add the condition `installedVersion != ""`, because we don't need to check this if the package operator is not installed yet.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,30 @@ var bootstrapCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg, _ := cliutils.RequireConfig(config.Kubeconfig)
 		client := bootstrap.NewBootstrapClient(cfg)
+
+		installedVersion, err := clientutils.GetPackageOperatorVersion(cmd.Context())
+		if err != nil {
+			IsBootstrapped, err := bootstrap.IsBootstrapped(cmd.Context(), cfg)
+			if err != nil && !IsBootstrapped {
+				fmt.Printf("error : %v\n", err)
+				return
```
christophenne, line 42: use `cliutils.ExitWithError()` instead of `return`

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,30 @@ var bootstrapCmd = &cobra.Command{
...
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = bootstrapCmdOptions.url
+		}
+		if !semver.IsUpgradable(installedVersion, desiredVersion) {
+			if !cliutils.YesNoPrompt("Do you want to downgrade the package-operator", false) {
```
christophenne, line 54: In this prompt, please also print the information about the version conflict, it should look like this:

"Glasskube is already installed in this cluster in the newer version %v. You are about to install version %v. This could lead to a broken cluster! 
Are you sure that you want to downgrade glasskube in this cluster?"

christophenne (changes requested):
@hanshal101 please have a look at the comments again, thanks!

hanshal101:
Here I have made the necessary changes, but when I bootstrap from `v0.3.0` to `v0.2.1` via `--url` it doesn't show the YesNo Prompt.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +32,30 @@ var bootstrapCmd = &cobra.Command{
...
+			}
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = ""
+		}
+		if !cliutils.YesNoPrompt(fmt.Sprintf("Glasskube is already installed in this cluster "+
```
christophenne, line 51: @hanshal101 where did the `semver.IsUpgradable` part go? the prompt should only be visible if `installedVersion != "" && !semver.IsUpgradable(...)`

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -31,6 +33,33 @@ var bootstrapCmd = &cobra.Command{
...
+			}
+		}
+
+		var desiredVersion string
+		if bootstrapCmdOptions.url == "" {
+			desiredVersion = config.Version
+		} else {
+			desiredVersion = ""
+		}
+		if !semver.IsUpgradable(installedVersion, desiredVersion) && installedVersion != "" {
```
christophenne, line 52: Sorry, I realized one more thing: If `installedVersion` and `desiredVersion` are the same, the prompt also appears, this should not happen. Because it's fine to bootstrap v0.3.0 again in a cluster where v0.3.0 is already running. 

So I think this should be fixed in this `if` by adding `&& installedVersion[1:] != desiredVersion`.

hanshal101:
Rebased && Pushed

christophenne (approved):
Thanks, please also squash all of your commits for the final merge.

hanshal101:
I hope this looks good!

PR #559, opened 2024-04-23
fix: add validating existing install before bootstrap

### LINKED ISSUE #454: `glasskube bootstrap` does not always apply all changes
**Describe the bug**
The `bootstrap` command does not always overwrite / reset all Glasskube resources, which lead to an inconsistent state. 

**To Reproduce**
1. Use a cluster for local development and execute  `make make install webhook cert`
2. Execute `glasskube bootstrap` 
3. See that the service type of the webhook service is still `External`

**Expected behavior**
Glasskube resources should be applied as defined

 - Type: Kubernetes
 - Version: 1.28
 - Provider: minikube

 - Package Operator version: v0.1.0
 - CLI version: v0.1.0

**Additional context**
Server side apply probably will not update fields it's not the owner of.

kosmoz:
Closes #454 

## 📑 Description
This PR adds a validation step to the bootstrap command: If any resource has the "kubectl last applied configuration" annotation, the operation is cancelled and no changes are made. This can be overridden using the new `--force` flag.

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -15,6 +15,7 @@ type bootstrapOptions struct {
 	bootstrapType    bootstrap.BootstrapType
 	latest           bool
 	disableTelemetry bool
+	force            bool
 }
 
 var bootstrapCmdOptions = bootstrapOptions{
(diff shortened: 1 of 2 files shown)


pmig (approved):

christophenne (approved):

pmig (approved):
LGTM

side note: job cannot be created if changed (maybe delete old jobs via bootstrap)

PR #565, opened 2024-04-24
feat(cli): cli version completion

### LINKED ISSUE #462: CLI version completion when updating to specific version

hanshal101:
Closes #462 
Fixes #462 

## 📑 Description

This PR add a new feature of auto completing the version for the update package command. 

## Method

Here I have added a new function for the completion of the available version. I have used the `describe.DescribePackage` function to identify the current version of the package installed. 

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -125,7 +125,8 @@ func completeUpgradablePackageVersions(
 	toComplete string,
 ) ([]string, cobra.ShellCompDirective) {
 
-	var dir cobra.ShellCompDirective
+	dir := cobra.ShellCompDirectiveNoFileComp
+
 	config, _, err := kubeconfig.New(config.Kubeconfig)
 	if err != nil {
 		dir &= cobra.ShellCompDirectiveError
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -10,7 +10,10 @@ import (
 	"github.com/glasskube/glasskube/api/v1alpha1"
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/config"
+	"github.com/glasskube/glasskube/internal/repo"
+	"github.com/glasskube/glasskube/internal/semver"
 	"github.com/glasskube/glasskube/pkg/client"
+	"github.com/glasskube/glasskube/pkg/describe"
 	"github.com/glasskube/glasskube/pkg/kubeconfig"
 	"github.com/glasskube/glasskube/pkg/statuswriter"
 	"github.com/glasskube/glasskube/pkg/update"
@@ -115,6 +118,34 @@ func completeInstalledPackageNames(
 		}
 	}
 	return
+}
+
+func completeUpgradablePackageVersions(
+	cmd *cobra.Command,
+	args []string,
+	toComplete string,
+) ([]string, cobra.ShellCompDirective) {
+	if len(args) == 0 {
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+	packageName := args[0]
+	var packageIndex repo.PackageIndex
+	if err := repo.FetchPackageIndex("", packageName, &packageIndex); err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+	_, _, _, installedVersion, err := describe.DescribePackage(cmd.Context(), packageName)
+	if err != nil {
+		fmt.Printf("error : %v", err)
+	}
+	versions := make([]string, 0, len(packageIndex.Versions))
+	for _, version := range packageIndex.Versions {
+		if toComplete == "" || strings.HasPrefix(version.Version, toComplete) {
+			if !semver.IsUpgradable(installedVersion, version.Version) {
+				versions = append(versions, version.Version)
+			}
+		}
+	}
+	return versions, cobra.ShellCompDirectiveNoFileComp
 }
 
 func init() {
--- a/cmd/glasskube/cmd/update.go
+++ b/cmd/glasskube/cmd/update.go
@@ -128,12 +128,12 @@ func completeUpgradablePackageVersions(
 	var dir cobra.ShellCompDirective
 	config, _, err := kubeconfig.New(config.Kubeconfig)
 	if err != nil {
-		dir = cobra.ShellCompDirectiveError
+		dir &= cobra.ShellCompDirectiveError
 		return nil, dir
 	}
 	client, err := client.New(config)
 	if err != nil {
-		dir = cobra.ShellCompDirectiveError
+		dir &= cobra.ShellCompDirectiveError
 		return nil, dir
 	}
 	if len(args) == 0 {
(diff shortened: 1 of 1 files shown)


kosmoz (changes requested):

hanshal101:
Hey @kosmoz I have made some changes, but still have a doubt in 2nd resolve which you gave. You can check this

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,39 @@ func completeInstalledPackageNames(
...
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+	packageName := args[0]
+	var packageIndex repo.PackageIndex
+	if err := repo.FetchPackageIndex("", packageName, &packageIndex); err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+	var pkg v1alpha1.Package
+	if err := client.FromContext(cmd.Context()).Packages().Get(cmd.Context(), packageName, &pkg); err != nil {
+		fmt.Println("error : ", err)
```
kosmoz, line 137: This print statement is unnecessary in the completion function.

kosmoz (changes requested):
Sorry @hanshal101, I made a mistake in my previous review: `client.FromContext` doesn't work in completion functions, because the context is not populated at that point. You can take a look at `completeInstalledPackageNames` (right above your added function) for reference how to actually get a working client. 

Again, this is totally my bad, sorry :upside_down_face: 

Also, I found another small thing please take a look below.

hanshal101:
Hey @kosmoz I have made some changes according to the `completeInstalledPackageNames` , please do review this and at last I'll rebase and refactor the code.

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
 	return
 }
 
+func completeUpgradablePackageVersions(
+	cmd *cobra.Command,
+	args []string,
+	toComplete string,
+) ([]string, cobra.ShellCompDirective) {
+
+	var dir cobra.ShellCompDirective
```
kosmoz, line 128: I think it would be good to initialize dir with `cobra.ShellCompDirectiveNoFileComp`.

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
...
+	config, _, err := kubeconfig.New(config.Kubeconfig)
+	if err != nil {
+		dir = cobra.ShellCompDirectiveError
```
kosmoz, line 131: Multiple directives can be combined, you already do this at line 152 (see below). I think this should be done everywhere consistently.

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
...
+		return nil, dir
+	}
+	client, err := client.New(config)
+	if err != nil {
+		dir = cobra.ShellCompDirectiveError
```
kosmoz, line 136: See previous comment

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
...
+	}
+	var pkg v1alpha1.Package
+	if err := client.Packages().Get(cmd.Context(), packageName, &pkg); err != nil {
+		dir &= cobra.ShellCompDirectiveError
+		return nil, dir
+	}
+	versions := make([]string, 0, len(packageIndex.Versions))
+	for _, version := range packageIndex.Versions {
+		if toComplete == "" || strings.HasPrefix(version.Version, toComplete) {
+			if !semver.IsUpgradable(pkg.Spec.PackageInfo.Version, version.Version) {
```
kosmoz, line 158: The negation here is incorrect. The current implementation suggests all versions that can NOT be upgraded to, but we obviously need the opposite :wink:

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
...
+		dir = cobra.ShellCompDirectiveError
+		return nil, dir
+	}
+	if len(args) == 0 {
+		return nil, dir
+	}
+	packageName := args[0]
+	var packageIndex repo.PackageIndex
+	if err := repo.FetchPackageIndex("", packageName, &packageIndex); err != nil {
+		return nil, dir
```
kosmoz, line 145: should also set `dir &= cobra.ShellCompDirectiveError` here

cmd/glasskube/cmd/update.go
```diff
@@ -117,9 +119,54 @@ func completeInstalledPackageNames(
...
+	}
+	if len(args) == 0 {
```
kosmoz, line 147: I think this condition could be extended to `len(args) != 1`. The reason is that the version flag should not be used when more than one package is upgraded.

kosmoz (changes requested):
The function works now! Well, in the sense that some versions are shown on completion :slightly_smiling_face: I still found some things that would need to be addressed but I think once these are done this PR should be ready to merge.

cmd/glasskube/cmd/update.go
hanshal101: Here, There is no such `cobra.ShellCompDirectiveNoFileComp`

hanshal101:
Also here, I think the `len(args)` is declared twice 
ref : [1](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R147) & [2](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R139)

cmd/glasskube/cmd/update.go
kosmoz: Mabye your IDE only completes type names? `dir := cobra.ShellCompDirectiveNoFileComp` works for me

hanshal101:
Well @kosmoz I have made all the changes here, also please have a look on this. 

> Also here, I think the `len(args)` is declared twice ref : [1](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R147) & [2](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R139)

kosmoz:
> Well @kosmoz I have made all the changes here, also please have a look on this.
> 
> > Also here, I think the `len(args)` is declared twice ref : [1](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R147) & [2](https://github.com/glasskube/glasskube/pull/565/files#diff-6eea45c4e741ad33659d1698862bc7850ee7f2c3582f0eae22fb548e1cef4035R139)

You are right, of course. I think the first one should be changed to `len(args) != 1` and the second one can be removed.

hanshal101:
Pushed all the changes 👍

kosmoz (approved):

PR #578, opened 2024-04-29
refactor(ui): switch to SSE and fix race conditions

### LINKED ISSUE #533: UI: Details page incomplete after uninstall completed
**Describe the bug**
When uninstalling a package, when the package has been deleted from the cluster the buttons on the top of the page disappear but the "install" part is still missing. After refreshing the page, it is there.

**To Reproduce**
1. Run the UI with `glasskube serve`
2. Go to an installed package
3. Press the uninstall button
4. Wait for completion

**Expected behavior**
the install form appears at the bottom of the page.

**Screenshots**
image

 - Type: Kubernetes
 - Version: v1.28.3

 - Package Operator version: 0.2.0
 - CLI version: 0.2.0

**Additional context**

### LINKED ISSUE #551: Uninstalling via UI does not show pending in OVH cluster
**Describe the bug**
While a package is being uninstalled, it is expected that the disabled Pending button is shown (although this should change with #456 – the problem would most likely still be the same). This works fine in minikube clusters, but seems to be different in a proper cluster (like in this case, OVH). Here the Pending button is shown for a very short time, and then it switches back to the Open/Uninstall button again. When uninstallation is finished, these buttons are removed correctly. 

**To Reproduce**
1. Uninstall a package via UI

**Expected behavior**
Pending button is shown during uninstallation. 

 - Type: Kubernetes
 - Version: 1.27.4-0
 - Provider: OVH

 - Package Operator version: v0.2.1
 - CLI version: v0.2.1

**Additional context**

* The events/updates arrive in the correct order at the client via the websocket, so it must be some client-side issue. 
* The issue does not arise when using the dev build or a locally built binary, only when using the released build.

christophenne:
Fixes #533 
Fixes #551 

## 📑 Description
In this PR we switch from using a websocket to the more lightweight server sent events (SSE), firstly because we only ever used the server-side push aspect of the websocket, and secondly because the [htmx integration](https://htmx.org/extensions/server-sent-events/) is nicer, as it allows to send down a trigger (e.g. "refresh-pkg-detail-cyclops") instead of fully rendered html. This way the client can "decide" whether to refetch some page or not. If no client is no an affected page currently, then the page will not even be rendered at all. As the refactoring shows, it also results in less complexity in how to render which template. 

## ✅ Checks


internal/web/ssehub.go
```diff
@@ -0,0 +1,91 @@
...
+	send chan *sse
+}
+
+// NewHub creates new SSEHub
+func NewHub() *SSEHub {
+	return &SSEHub{
+		Broadcast:  make(chan *sse),
+		Register:   make(chan *sseClient),
+		Unregister: make(chan *sseClient),
+		clients:    make(map[*sseClient]bool),
```
kosmoz, line 44: Does the `bool` do anything here? If not, you could use a `map[*sseClient]struct{}`, which is a little bit more efficient.

kosmoz (commented):
In addition to what we discussed, here are two minor improvements I could see

kosmoz (approved):

Issue #207, opened 2024-02-06
Add user feedback to packages


pmig:
**Is your feature request related to a problem? Please describe.**
Package managers like the App Store, Play Store or [Arch User Repository](https://aur.archlinux.org/) provide a possibility to comment or upvote a package.

**Describe the solution you'd like**
We could start with a discussion threads for every package on the packages repo which we then could either link to it from the website or the GUI or get these comments via the GitHub API.

**Additional context**
This is not something we would start immediately but more a ticket the will get more important in the longrun.

kosmoz:
Maybe also giscus?

Related: https://github.com/glasskube/glasskube/issues/439

christophenne:
Discussion page has been added in #596. 
Follow-up TODOs:
[ ] make pkg-detail-header show back button when on discussion page
[ ] show number of reactions and comments (using bootstrap icons – something like chat-left-heart and chat-left-text) instead of "Discussion" link on the package detail header
[ ] show number of reactions/comments (maybe summed up "number of interactivity") for each package on the package overview page.

pmig:
It's already done, yes

christophenne:
> * [ ] show number of reactions/comments (maybe summed up "number of interactivity") for each package on the package overview page

was still open, but if we need this we can create it again

kosmoz:
> > * [ ]  show number of reactions/comments (maybe summed up "number of interactivity") for each package on the package overview page
> 
> was still open, but if we need this we can create it again

Yeah, but TBH I don't like this feature. Let's create a new issue if we decide that we want it though.

PR #596, opened 2024-05-03
feat(ui): integrate giscus for package feedback (#207)

christophenne:
Ref #207

## 📑 Description
This PR introduces a new discussion page on for each package, showing the giscus component.

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -150,6 +150,7 @@ func (s *server) Start(ctx context.Context) error {
 	router.Handle("/packages/uninstall/modal", s.requireReady(s.uninstallModal))
 	router.Handle("/packages/open", s.requireReady(s.open))
 	router.Handle("/packages/{pkgName}", s.requireReady(s.packageDetail))
+	router.Handle("/packages/{pkgName}/discussion", s.requireReady(s.packageDiscussion))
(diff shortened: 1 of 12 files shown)


internal/web/server.go
```diff
@@ -549,7 +581,7 @@ func (s *server) kubeconfigPage(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *server) enrichWithErrorAndWarnings(ctx context.Context, data map[string]any, err error) map[string]any {
-	cacheBustStr := fmt.Sprintf("%d", rand.Int())
+	cacheBustStr := config.Version // fmt.Sprintf("%d", rand.Int())
```
kosmoz, line 584: Please remove this commented out code. Additionally, with this change, I think setting the value again below is redundant, since the clientVersion is the same as config.Version.

internal/web/root/static/custom.js
kosmoz, line 1: Maybe we should move this file to `/web` and create a bundle for it?

kosmoz (changes requested):
Could you please take a look a the comments below? Apart from those, everything works great. 

Theme switching also worked nicely for me

christophenne:
> Theme switching also worked nicely for me

Turns out this weird bug only appears in chrome :smiley:

kosmoz (approved):

Issue #491, opened 2024-04-09
Handle bootstrap errors gracefully when glasskube is not connected to the internet


Baalekshan:
**Is your feature request related to a problem? Please describe.**

When the user is not connected to the internet and `glasskube bootstrap` is used the bootstrap fails and returns a memory invalid or nil pointer deference error but instead we should print a message as glasskube is not connected to the internet. Something like this

```
~$ glasskube bootstrap
⚠️ Glasskube is not connected to the internet, cannot bootstrap.
```

**Describe the solution you'd like**
Improve error handling for `cmd/bootstrap.go`

Addition from @christophenne: I think the error message doesn't necessarily have to be "not connected to internet", but can be more generic. The point where the first call to some http endpoint is made, is when fetching the lastest version of glasskube. It should be fine to do the same error handling as used in `clientutils.FetchResources`.

**Describe alternatives you've considered**

**Additional context**

Baalekshan:
@christophenne this could be a good-first-issue

christophenne:
@Baalekshan good point, but I think this is being solved already with #524, right?

Baalekshan:
No that's a different issue involving --url flag.
This one is for showing that glasskube is not connected to the internet when tried bootstrapping.This one might be status code 503 service unavailable when device is offline. #524 deals with 404 error.

christophenne:
@Baalekshan fair enough, good input.

Aryakoste:
Hello @christophenne. I would like to work on this issue.

PR #597, opened 2024-05-03
refactor: added proper error handling for network related errors

Aryakoste:
Closes https://github.com/glasskube/glasskube/issues/491

## 📑 Description

When there is network timeout or the device is offline while using glasskube bootstrap appropriate error will be printed in the terminal.

## ✅ Checks

### DIFF
--- a/internal/httperror/error.go
+++ b/internal/httperror/error.go
@@ -28,11 +28,13 @@ func IsErrorResponse(respose *http.Response) bool {
 	return respose.StatusCode >= 400
 }
 
-func CheckResponse(response *http.Response) error {
-	if IsErrorResponse(response) {
-		return &statusError{response.Status, response.StatusCode}
+func CheckResponse(response *http.Response, err error) (*http.Response, error) {
+	if err != nil {
+		return response, err
+	} else if IsErrorResponse(response) {
+		return response, &statusError{response.Status, response.StatusCode}
 	} else {
-		return nil
+		return response, nil
 	}
 }
 
@@ -49,7 +51,7 @@ func IsNotFound(err error) bool {
 	return Is(err, http.StatusNotFound)
 }
 
-func IsNetworkError(err error) bool {
-	netErr, ok := err.(net.Error)
-	return ok && netErr.Timeout()
+func IsTimeoutError(err error) bool {
+	var netErr net.Error
+	return errors.As(err, &netErr) && netErr.Timeout()
 }
--- a/internal/httperror/error.go
+++ b/internal/httperror/error.go
@@ -3,6 +3,7 @@ package httperror
 import (
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 )
 
@@ -46,4 +47,9 @@ func Is(err error, code int) bool {
 
 func IsNotFound(err error) bool {
 	return Is(err, http.StatusNotFound)
+}
+
+func IsNetworkError(err error) bool {
+	netErr, ok := err.(net.Error)
+	return ok && netErr.Timeout()
 }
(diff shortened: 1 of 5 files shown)


internal/httperror/error.go
```diff
@@ -47,3 +48,8 @@ func Is(err error, code int) bool {
 func IsNotFound(err error) bool {
 	return Is(err, http.StatusNotFound)
 }
+
+func IsNetworkError(err error) bool {
+	netErr, ok := err.(net.Error)
+	return ok && netErr.Timeout()
+}
```
kosmoz, line 55: Two things here:
1. I think `IsTimeoutError` would be a more fitting name
2. Plase use [`errors.As`](https://pkg.go.dev/errors#As) instead of a simple cast. It is more resilient and also checks wrapped errors. Something like this should work:
```go
var netErr net.Error 
return errors.As(err, &netErr) && netErr.Timeout()
```

internal/releaseinfo/releaseinfo.go
```diff
@@ -12,8 +15,20 @@ type ReleaseInfo struct {
...
-	resp, _ := http.Get(url)
-
+	resp, err := http.Get(url)
+	if err != nil {
+		if err := httperror.CheckResponse(resp); err != nil {
+			switch {
+			case httperror.IsNetworkError(err):
+				return nil, fmt.Errorf("Network Timeout, check your network %v: %v", url, err)
+			case httperror.Is(err, http.StatusServiceUnavailable):
+				return nil, fmt.Errorf("Could not connect to the glasskube server. Please check your internet connection and try again. %v: %v", url, err)
```
kosmoz, line 25: Since this is a relatively low-level function, I think there should be no calls to action (e.g. "check your network") in the error messages. They can be added higher up where this function is used.

kosmoz (changes requested):
Thanks for your PR @Aryakoste! :rocket: 

I tried your changes with network disabled and unfortunately I got a SIGSEGV. Please see my comments below.

internal/releaseinfo/releaseinfo.go
```diff
@@ -12,8 +15,20 @@ type ReleaseInfo struct {
...
+			default:
+				return nil, fmt.Errorf("Failed to fetch glasskube version %v: %v", url, err)
+			}
+		}
+		return nil, err
+	}
```
kosmoz, line 31: Unfortunately, I think there are some mistakes here. This is understandable because network errors in Go can be a little confusing…

1. If `err` is initially not `nil`, this means that the HTTP request has failed and the response object should not be used (it is `nil`). In this case this leads to a SIGSEGV because `httperror.CheckResponse` requires that the response is not `nil`.
2. `httperror.CheckResponse` does not return a `net.Error`, so calling `httperror.IsNetworkError` on the error returned by it can never be true.

I can think of two ways to improve this: 
1. Locally, by checking `err` first and then handling a possible HTTP error
```go
if err != nil {
  /* check network error here */
} else if err := httperror.CheckResponse(resp); err != nil {
  /* handle HTTP error here */
}
```
2. Globally, by changing `httperror.CheckResponse` to accept and return both an error and a response: 
```go
func CheckResponse(response *http.Response, err error) (*http.Response, error) {
	if err != nil {
		return response, err
	} else if IsErrorResponse(response) {
		return response, &statusError{response.Status, response.StatusCode}
	} else {
		return response, nil
	}
}
```

This is very neat because it can be used like this:

```go
resp, err := httperror.CheckResponse(http.Get(url))
```

I think option (2) would be preferable, but if you rather want to do option (1) that's fine by me too.

kosmoz (approved):

PR #609, opened 2024-05-06
fix(cli): version prefix v now is now optional

### LINKED ISSUE #463: Handle `v` prefix in versions in install and update command
`install` and `update` commands offer the possibility to update to a specific verison with the `-v` flag. The version entered must have the `v` prefix, or otherwise the package will not be found in the package repo. That's not the greatest user experience.

Example:
`glasskube install keptn -v v2.0.0-rc.1+1` works, because the version number has the `v` prefix.
`glasskube install keptn -v 2.0.0-rc.1+1` does not work, because the version number is missing the `v`.

The user should be able to enter their version without the `v` prefix. This should work for the install command and also for the update command. 

**Additional context**
#412 is still in progress.

sumeshmurali:
Closes #463

## 📑 Description
The PR make changes to `-v/--version` in the glasskube cli. Previously it requires the user to add "v" prefix to the version. With the PR the user will be able to pass exact version and the code internally adds a "v" prefix. 

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -64,7 +64,7 @@ var installCmd = &cobra.Command{
 			installCmdOptions.Version = packageIndex.LatestVersion
 			fmt.Fprintf(os.Stderr, "Version not specified. The latest version %v of %v will be installed.\n",
 				installCmdOptions.Version, packageName)
-		} else if installCmdOptions.Version[:1] != "v" { // the slice notation is used to get a string instead of byte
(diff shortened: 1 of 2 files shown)


kosmoz (changes requested):
Thanks @sumeshmurali!

This looks good, but I think it would be a little cleaner to use `strings.HasPrefix` instead. Could you please make this change?

sumeshmurali:
> This looks good, but I think it would be a little cleaner to use `strings.HasPrefix` instead. Could you please make this change?

Made the changes. Can you check now?

kosmoz (approved):

PR #619, opened 2024-05-08
fix(cli): error updating using bootstrap

### LINKED ISSUE #613: Error updating using `bootstrap`
**Describe the bug**
When using `glasskube bootstrap` to upgrade the cluster components from version 0.2 to 0.3 there is an error because the webhook cert job cannot be updated.

**To Reproduce**
1. Setup a fresh cluster
2. Bootstrap an older version
3. Bootstrap a newer version

**Expected behavior**
If a job with the same name already exists, do one of the following:
- It is not updated
- It is deleted and re-created using the updated manifest

This should not be limited to the webhook cert job, it should work for any potential job we might add.

**Additional context**

hanshal101:
Closes #613 
Fixes #613 

## 📑 Description

This PR fixes the error while running the bootstrap command. When upgrading the package-operator the `jobs` would return back as there was a conflict in the names (i.e the desired and the existing job names were similiar).

## Solution
Made a checkpoint in the `applyManifest` function where if the Resource type is `Job` then first it will delete the existing one and then Create the desired one.

## ✅ Checks

## ℹ Additional Information

I first though of applying this in the `preprocessManifests` but removing the `Job` here would led to conflicts in the `Apply` of `applyManifest` where it will again try to update the Job which will again return an error.


hanshal101:
If this looks good to you I'll rebase and push!

pkg/bootstrap/bootstrap.go
```diff
@@ -188,14 +188,30 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
 		bar.Describe(fmt.Sprintf("Applying %v (%v)", obj.GetName(), obj.GetKind()))
-		if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
-				Apply(ctx, obj.GetName(), &obj, metav1.ApplyOptions{Force: true, FieldManager: "glasskube"})
-			return err
-		}); err != nil {
-			return err
+		if obj.GetKind() == "Job" {
+			options := metav1.DeletePropagationBackground
+			fmt.Println("Deleting Job")
```
kosmoz, line 193: Please remove this print statement, it doesn't work with the progress bar.

pkg/bootstrap/bootstrap.go
```diff
@@ -188,14 +188,30 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
+				metav1.DeleteOptions{PropagationPolicy: &options})
+			if err != nil {
+				return err
+			}
+			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Create(ctx, &obj, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+		} else {
+			fmt.Println("On : ", obj.GetKind())
```
kosmoz, line 206: Please remove this print statement, it doesn't work with the progress bar.

pkg/bootstrap/bootstrap.go
```diff
@@ -188,14 +188,30 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
+			options := metav1.DeletePropagationBackground
+			fmt.Println("Deleting Job")
+			err := c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Delete(
+				ctx,
+				obj.GetName(),
+				metav1.DeleteOptions{PropagationPolicy: &options})
+			if err != nil {
+				return err
+			}
+			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Create(ctx, &obj, metav1.CreateOptions{})
```
kosmoz, line 201: I think we should use `Apply` instead of create here as well, since this is also done in the `else` branch.

kosmoz (changes requested):
Thanks @hanshal101! Please see my comments below. And yes, please rebase your branch as well :slightly_smiling_face:

pkg/bootstrap/bootstrap.go
hanshal101: Oh, but `Apply` just updates the current resources instead of creating them?

kosmoz: It can do both :slightly_smiling_face:

pkg/bootstrap/bootstrap.go
```diff
@@ -188,14 +188,32 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
-			return err
+		if obj.GetKind() == "Job" {
+			options := metav1.DeletePropagationBackground
+			err := c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).Delete(
+				ctx,
+				obj.GetName(),
+				metav1.DeleteOptions{PropagationPolicy: &options})
+			if err != nil {
+				return err
+			}
```
kosmoz, line 199: There is an error here if the cluster is not yet bootstrapped. We need to ignore NotFound errors. The canonical way to do this is to replace `if err != nil` with `if err != nil && !apierrors.IsNotFound(err)`.

pkg/bootstrap/bootstrap.go
```diff
@@ -188,14 +188,32 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
+		} else {
+			if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
+					Apply(ctx, obj.GetName(), &obj, metav1.ApplyOptions{Force: true, FieldManager: "glasskube"})
+				return err
+			}); err != nil {
+				return err
+			}
 		}
```
kosmoz, line 216: This part is basically the same as the `else` branch. It would be better to remove this part and lift the code out of the else branch.

Something like this:

```suggestion
			}
		}

		if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
				Apply(ctx, obj.GetName(), &obj, metav1.ApplyOptions{Force: true, FieldManager: "glasskube"})
			return err
		}); err != nil {
			return err
		}

```

kosmoz (changes requested):

kosmoz (approved):

PR #638, opened 2024-05-14
feat(cli): add json and yaml format in the list sub-command

### LINKED ISSUE #465: JSON output for list
Hey folks,

doing integration of your app in my solution and faced one thing - lack of JSON output for `glasskube list`, want something like `... --format json` like in Helm.

Thanks.

### LINKED ISSUE #629: JSON/YAML output for CLI commands
All CLI commands should support specifying an output format if applicable. This can be done using the `--output` flag (short: `-o`). This flag is optional (default: empty) and can be set to "json" or "yaml". Output should be written to STDOUT.

If a command modifies some resources, the output should be of the resource after modification. 
Those commands should also provide a `--dry-run` option. 

### Examples
```
# list packages as yaml
glasskube ls -o yaml
# simulate install and write json output to a file
glasskube install foo-pkg --dry-run --output=json > foo.json
```

### Subtasks
- [x] https://github.com/glasskube/glasskube/issues/465
- [x] #655 
- [x] #656
- [x] #659
- [x] #657
- [x] #660
- [x] #658
- [x] #661
- [x] #724
- [x] #725

hanshal101:
Closes #465 
Fixes #465

## 📑 Description

This PR adds the feature of printing the list subcommand in the `json` format.

## ✅ Checks

## ℹ Additional Information
In this code I tried the `JSON.Marshal()` function but the output was not quite readable, hence went with the approach of 
`JSON.MarshalIndent()`.

### DIFF
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -157,18 +157,15 @@ func printPackageJSON(packages []*list.PackageWithStatus) {
 }
 
 func printPackageYAML(packages []*list.PackageWithStatus) {
-	var isFirstPackage = true
 
-	for _, pkg := range packages {
+	for i, pkg := range packages {
 		yamlData, err := yaml.Marshal(pkg)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "error marshaling data to YAML: %v\n", err)
 			cliutils.ExitWithError()
 		}
 
-		if isFirstPackage {
-			isFirstPackage = false
-		} else {
+		if i > 0 {
 			fmt.Println("---")
 		}
 
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -19,6 +20,7 @@ type ListCmdOptions struct {
 	ShowDescription   bool
 	ShowLatestVersion bool
 	More              bool
+	ListFormat        string
 }
 
 func (o ListCmdOptions) toListOptions() list.ListOptions {
@@ -59,8 +61,13 @@ var listCmd = &cobra.Command{
 				fmt.Fprintln(os.Stderr, "No packages found. This is probably a bug.")
 			}
 		} else {
-			printPackageTable(pkgs)
+			if listCmdOptions.ListFormat == "json" {
+				printPackageJSON(pkgs)
+			} else {
+				printPackageTable(pkgs)
+			}
 		}
+
 	},
 }
 
@@ -75,6 +82,7 @@ func init() {
 		"show the latest version of packages if available")
 	listCmd.PersistentFlags().BoolVarP(&listCmdOptions.More, "more", "m", false,
 		"show additional information about packages (like --show-description --show-latest)")
+	listCmd.PersistentFlags().StringVar(&listCmdOptions.ListFormat, "format", "", "output format (json, yaml, etc.)")
 
 	listCmd.MarkFlagsMutuallyExclusive("show-description", "more")
 	listCmd.MarkFlagsMutuallyExclusive("show-latest", "more")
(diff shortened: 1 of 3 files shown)


cmd/glasskube/cmd/list.go
```diff
@@ -19,6 +20,7 @@ type ListCmdOptions struct {
 	ShowDescription   bool
 	ShowLatestVersion bool
 	More              bool
+	ListFormat        string
```
kosmoz, line 23: Cobra (the CLI library we use) supports custom types for flags by adding `String()` `Set()` and `Type()` functions to a type. I think it would make sense to use this feature here, since the same flag will be added to multiple commands in the future. 

We do something similar with the bootstrap type, which you can look at for reference: https://github.com/glasskube/glasskube/blob/main/pkg/bootstrap/type.go

Let me know if you have any questions regarding this.

kosmoz (changes requested):
Nice work @hanshal101! Do you plan to implement YAML formatting in the same PR? I think it would make sense, then we can close the issue with a single PR.

cmd/glasskube/cmd/list.go
hanshal101: Well here in the `init` function should have the `VarP` instead of `StringVarP`
Because I took the reference to the Bootstrap Type command and it has `VarP`
https://github.com/glasskube/glasskube/blob/c2d0fcf4306aa9c9f5c716143d644004e6a5f646/cmd/glasskube/cmd/bootstrap.go#L87

kosmoz: That's right. 

First, you add a custom string type. 
```go
type OutputFormat string
```
Then you define possible values.

```go
var (
  OutputFormatJSON OutputFormat = "json"
  OutputFormatYAML OutputFormat = "yaml"
)
```

Next, you add the required functions as mentioned and change the option type and flag declaration as you described.

cmd/glasskube/cmd/list.go
```diff
@@ -109,6 +146,28 @@ func printPackageTable(packages []*list.PackageWithStatus) {
...
+
+func printPackageYAML(packages []*list.PackageWithStatus) {
+	for _, pkg := range packages {
+		yamlData, err := yaml.Marshal(pkg)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error marshaling data to YAML: %v\n", err)
+			cliutils.ExitWithError()
+		}
+		fmt.Println(string(yamlData))
+		fmt.Println("---")
```
kosmoz, line 167: This produces an empty document at the end. The boundary marker should be printed before the `yamlData`. Additionally, I think it would make sense to only print the boundary marker if there are at least two packages.

kosmoz (changes requested):
Works nicely! In addition to the one commend below, I noticed that the output has some mixed casing for the JSON/YAML keys. Can you please add some JSON tags to the `PackageWithStatus` struct at `/pkg/list/list.go`? For example, `Status` should be converted to `status`, so this tag would be appropriate `json:"status,omitempty"`

hanshal101:
Well here I have added a small if else to see that if the current package is the first package or not and moved the `---` above the packages. This will help to eliminate the first package and print in rest others. How's this !

cmd/glasskube/cmd/list.go
```diff
@@ -109,6 +146,36 @@ func printPackageTable(packages []*list.PackageWithStatus) {
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "error marshaling data to YAML: %v\n", err)
+			cliutils.ExitWithError()
+		}
+
+		if isFirstPackage {
+			isFirstPackage = false
+		} else {
+			fmt.Println("---")
+		}
```
kosmoz, line 173: I think this could be simplified by utilizing the loop counter instead, like:
```go
for i, pkg := range packages {
    /* ... */
    if i > 0 {
        fmt.Println("---")
    }
    /* ... */
}
```

kosmoz (changes requested):
Almost perfect! Could you please address these two comments? Afterwards, I think this is ready to merge!

kosmoz (approved):

PR #643, opened 2024-05-15
feat(ui): add settings page with list of repositories

christophenne:
## ✅ Checks

### DIFF
--- a/internal/web/templates/layout/base.html
+++ b/internal/web/templates/layout/base.html
@@ -16,31 +16,30 @@
     sse-connect="/events"
     hx-indicator="#indicator"
     hx-target-error="#error-container">
-    <nav
-      hx-boost="true"
-      hx-select="main"
-      hx-target="main"
-      hx-swap="outerHTML"
-      class="navbar navbar-expand-lg navbar-dark bg-secondary">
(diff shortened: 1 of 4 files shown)


internal/web/templates/layout/base.html
```diff
@@ -16,31 +16,30 @@
...
-      hx-target="main"
-      hx-swap="outerHTML"
-      class="navbar navbar-expand-lg navbar-dark bg-secondary">
+    <nav class="navbar navbar-expand-lg navbar-dark bg-secondary">
       <div id="indicator" class="progress-container bg-transparent w-100 position-fixed top-0 start-0">
         <div class="htmx-indicator progress-bar bg-primary h-100 w-100"></div>
       </div>
 
-      <div class="container">
+      <div class="container" hx-swap-oob="outerHTML:#navbar-content" id="navbar-content">
```
christophenne, line 24: out-of-bounds swap is necessary, because usually at page changes we just swap the `main` element, which does not include the navbar. Another option would be to swap the whole `body`, but as far as I remember there where some issues then with the `htmx:historyRestore` trigger and therefore with going back and forth with the browser history buttons.

kosmoz (approved):

PR #644, opened 2024-05-15
chore: telemetry for repositories

christophenne:
## ✅ Checks

### DIFF
--- a/internal/telemetry/properties/properties.go
+++ b/internal/telemetry/properties/properties.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"strings"
 
+	"github.com/glasskube/glasskube/internal/constants"
+
+	"github.com/glasskube/glasskube/api/v1alpha1"
+
 	"github.com/glasskube/glasskube/internal/telemetry/annotations"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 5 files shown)


internal/telemetry/properties/properties.go
```diff
@@ -69,3 +82,20 @@ func (g PropertyGetter) ClusterProperties() (p ClusterProperties) {
...
+
+func (g PropertyGetter) RepositoryProperties() (p RepositoryProperties) {
+	if g.RepositoryLister != nil {
+		if ls, err := g.RepositoryLister.ListRepositories(context.Background()); err == nil {
+			p.nrepositories = len(ls.Items)
+			for _, repo := range ls.Items {
+				if repo.Spec.Auth != nil {
+					p.nrepositoriesAuth = p.nrepositoriesAuth + 1
+				}
+				if repo.IsDefaultRepository() && !strings.Contains(repo.Spec.Url, "https://packages.dl.glasskube.dev/packages") {
```
kosmoz, line 94: Just two minor things:
1. Maybe we should extract the default URL to a constant? Put it in `/internal/constants/`?
2. Is there a reason you use `string.Contains` vs a simple `!=`?

christophenne: ad 2.: I thought just to be safe about the trailing slash. Someone could tinker around and change the url or delete the repo, then change it back or recreate it and do it without the trailing slash. Granted, a very very specific case.

kosmoz: Fair enough, we can keep it that way

kosmoz (approved):

PR #649, opened 2024-05-17
feat(ui): show uninstalling button if a package is currently being uninstalled (#456)

### LINKED ISSUE #456: Show a different button state if a package is currently being uninstalled
Currently, a package is displayed as "pending" after starting an uninstall action, provided that the package-operator is running in the cluster. If no package-operator is running, the displayed state doesn't change until the package is finally removed after GC has completed (we use foreground deletion for Glasskube packages).

It would be better to detect if a package has been marked for deletion and display a different button state on the overview and details pages if it is. To start with, the button could look like the "pending" button but with a different text, e.g. "uninstalling". It is important that the button is also `disabled`.

A resource can be considered marked for deletion if it has a `DeletionTimestamp`:
```go
if !pkg.DeletionTimestamp.IsZero() {
  // pkg is being deleted ...
}
```

Baalekshan:
…installed (#456)

Closes #456 

## 📑 Description

Shows Uninstalling package button when the package is being uninstalled

On package overview page
image

On package details page
image

## ✅ Checks

### DIFF
--- a/internal/web/components/pkg_detail_btns/controller.go
+++ b/internal/web/components/pkg_detail_btns/controller.go
@@ -15,6 +15,7 @@ type pkgDetailBtnsInput struct {
 	Status          *client.PackageStatus
 	Manifest        *v1alpha1.PackageManifest
 	UpdateAvailable bool
+	Pkg             *v1alpha1.Package
 }
 
 func getId(pkgName string) string {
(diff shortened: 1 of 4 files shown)


christophenne (approved):
Good job, thanks!

PR #669, opened 2024-05-21
feat(cli): add `--output` flag for `glasskube update`

### LINKED ISSUE #657: Add `--output` support for `glasskube update`
See #629 for details.

Since `update` modifies a resource, the output of the command should be the updated package.

Aryakoste:
Closes #657 

## ✅ Checks


kosmoz (commented):
I would put this on hold until https://github.com/glasskube/glasskube/pull/670#pullrequestreview-2070528628 is resolved.

kosmoz (changes requested):
Hey @Aryakoste! Since your other PR just got merged, could you please update this one to use the new `OutputOptions`?

Aryakoste:
Done @kosmoz

kosmoz (changes requested):
As noted in the issue description, the output of the command should be the list of serialized `Package` resources after being updated, not the update transaction. 

To achieve this, you will probably have to modify the `updater.Apply` function to return that data. If the output is YAML, the list should be printed as multi-document YAML, similar to what we already do in the `ListCommand`.

Aryakoste:
Changes done check and let me know @kosmoz

cmd/glasskube/cmd/update.go
```diff
@@ -93,6 +99,41 @@ func printTransaction(tx update.UpdateTransaction) {
...
+	case OutputFormatYAML:
+		var buffer bytes.Buffer
+		for _, pkg := range pkgs {
+			data, err := yaml.Marshal(pkg)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "❌ failed to marshal output: %v\n", err)
+				cliutils.ExitWithError()
+			}
+			buffer.Write(data)
+			buffer.Write([]byte("---\n"))
```
kosmoz, line 121: This leads to an empty document at the end. Please print the document separator before the content and only if there are two or more packages.

kosmoz (changes requested):
Looks good now. Please have a look at this final comment, afterwards I think this is ready to merge!

Aryakoste:
@kosmoz Changes done.

kosmoz (changes requested):
Looks great but can you please ensure that all packages have the correct `TypeMeta`? With that addressed, I think this would be ready for merge.

You can check out https://github.com/Aryakoste/glasskube/blob/7b9ce5ba8346c3ad20c8e4b10d3e066f2dcbef26/cmd/glasskube/cmd/configure.go#L90-L92 as a reference

kosmoz (approved):

PR #670, opened 2024-05-21
Added --output support for glasskube configure

### LINKED ISSUE #658: Add `--output` support for `glasskube configure`
See #629 for details.

Since `configure` modifies a resource, the output of the command should be the configured package.

Aryakoste:
Closes #658 

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/output_options.go
+++ b/cmd/glasskube/cmd/output_options.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


kosmoz (changes requested):
Hi @Aryakoste, thanks for your PR! :rocket: 

This change looks very clean, but I'm afraid that we are implementing the same thing in 7 different ways at the moment. @hanshal101 already implemented something similar in #638 and I think it would make sense to extract at least the options part to make it reusable. We already have some relevant examples of this, which can be taken as a reference: https://github.com/glasskube/glasskube/blob/main/internal/manifestvalues/flags/flags.go

In this case, I would create a file `cmd/glasskube/cmd/output_options.go`. There I would move the `ListFormat` type, which should be renamed to `OutputFormat`. I would also add a `type outputOptions struct` that has an `Output` field of type `OutputFormat`. This type should also have a function like this:

https://github.com/glasskube/glasskube/blob/da30108d42431ed96845d69483b3f026f4342e66/internal/manifestvalues/flags/flags.go#L37-L51

Then the `OutputOptions` type can be used similar to this: 

https://github.com/glasskube/glasskube/blob/main/cmd/glasskube/cmd/install.go#L29

Do you think you could tackle this change? Since this is a somewhat advanced refactoring, I can also take this PR from here if you prefer. Let me know if you have any questions!

Aryakoste:
I can tackle this changes. Will work on it and let you know. @kosmoz

cmd/glasskube/cmd/output_options.go
```diff
@@ -0,0 +1,21 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type OutputFormat string
```
kosmoz, line 7: `OutputFormat` should implement [`pflag.Value`](https://pkg.go.dev/github.com/spf13/pflag@v1.0.5#Value). To do this, you have to create the `String()`, `Set(string)` and `Type()` functions. `ListFormat` already has those functions you can just copy them. 

You can then use `flags.VarP(...)` instead of `flags.StringVarP(...)` with an explicit cast. 

Finally, please adapt the list command to also use `OutputOptions` and remove the `ListFormat` type, since it should be no longer used by then.

Again, let me know if you have any questions or require assistance!

kosmoz (changes requested):

Aryakoste:
@kosmoz Changes done. Check and let me know.

kosmoz (approved):

Issue #496, opened 2024-04-10
Dealing with broken configuration references


christophenne:
**Is your feature request related to a problem? Please describe.**
A user can enter different kinds of references to choose config values for a package (e.g. ConfigMap, Secret, Package Config), both on the UI and CLI. They don't yet see whether the entered reference (e.g. namespace + name + key in the case of a ConfigMap) can be resolved and is valid (at least on the UI). 

**Describe the solution you'd like**
For a reference input, the user should see some hint of whether the entered reference is:
* existing and passes validation
* existing but does not pass validation
* not existing

**To be discussed**: This kind of hint should probably not be a strict validation? As also described in #495, when a referenced object gets changed or deleted, the configuration could become invalid. We need a way to deal with this. 

There are lots of ways in which a reference could become invalid:
* a referred value changes, and the new value is not valid according to the glasskube package anymore
* a referred value is deleted (e.g. because the ConfigMap, Secret or Package has been removed)

**Describe alternatives you've considered**

**Additional context**
Follow-Up of #121 
Related to #495

kosmoz:
Notably, the operator already deals with this by not touching any dependent resources and setting a failed condition on the Package resource. This ticket is only about how to handle this situation on the UI (and to a lesser extend CLI).

kosmoz:
https://github.com/glasskube/glasskube/pull/675 implements resolving reference values and showing potential errors. 

We still don't validate those resolved values.

PR #675, opened 2024-05-22
feat(ui): show broken config references (#496)

christophenne:
Ref #496

## ✅ Checks

### DIFF
--- a/internal/util/error.go
+++ b/internal/util/error.go
@@ -1,0 +1,12 @@
+package util
+
+import "errors"
+
+func GetRootCause(err error) error {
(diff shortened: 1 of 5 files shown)


internal/util/error.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import "errors"
+
+func GetDeepestError(err error) error {
```
kosmoz, line 5: I think `GetRootCause` would be a little easier to understand here

kosmoz (commented):
One small nit. Apart from that, looking good!

kosmoz (approved):

PR #689, opened 2024-05-24
feat(ui): autocomplete for reference inputs (#495)

### LINKED ISSUE #495: Search/Autocomplete in Package Configuration Page
On the package configuration page, the user can enter reference values (ConfigMap, Secret, Package Config), but the input fields are classic text fields without any additional support to show which options there are. 

For ConfigMap and Secret inputs, the namespace field should show the namespaces of the cluster, and if a namespace has been entered, the name field should show the config maps of this namespace. If possible, once the namespace + name have been entered, the key field shows the possible keys to select from. 

For Package inputs, the package field should show all installed packages, and if a package has been selected, the value field should show the possible values. 

All of the inputs should support having just plain text too (i.e. none of the available options): Even if we don't allow users to input non-existing objects (to be discussed), someone could always delete some config map or secret afterwards – but the reference should still be visible in the UI.

Classic select without text input field – easier to implement but more difficult for the user if there are a lot of items. Also has the downside that it cannot display broken references (e.g. a reference that has become invalid because a referred config map was deleted). 

**Additional context**
Follow-Up of #121 

The most basic way to do this would be an html datalist element, which has text filtering on statically defined options out of the box. But there are certainly more advanced ways to implement this too.

christophenne:
Closes #495

## ✅ Checks

### DIFF
--- a/internal/web/configurationvalues.go
+++ b/internal/web/configurationvalues.go
@@ -1,22 +1,42 @@
 package web
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"os"
+	"slices"
 	"strconv"
 	"strings"
+
+	"github.com/glasskube/glasskube/pkg/manifest"
+
+	"github.com/glasskube/glasskube/internal/maputils"
+	"github.com/glasskube/glasskube/internal/web/components/pkg_config_input"
+	"github.com/glasskube/glasskube/pkg/describe"
+	"github.com/gorilla/mux"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/glasskube/glasskube/api/v1alpha1"
 )
 
 const (
-	namespaceKey = "namespace"
-	nameKey      = "name"
-	keyKey       = "key"
-	packageKey   = "package"
-	valueKey     = "value"
-	refKindKey   = "refKind"
+	namespaceKey     = "namespace"
+	nameKey          = "name"
+	keyKey           = "key"
+	packageKey       = "package"
+	valueKey         = "value"
+	refKindKey       = "refKind"
+	refKindConfigMap = "ConfigMap"
+	refKindSecret    = "Secret"
+	refKindPackage   = "Package"
 )
+
+func formKey(valueName string, key string) string {
+	return fmt.Sprintf("%s[%s]", valueName, key)
+}
 
 func extractValues(r *http.Request, manifest *v1alpha1.PackageManifest) (map[string]v1alpha1.ValueConfiguration, error) {
 	values := make(map[string]v1alpha1.ValueConfiguration)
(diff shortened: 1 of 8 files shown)


internal/web/server.go
```diff
@@ -17,6 +17,11 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/glasskube/glasskube/internal/web/components/pkg_config_input"
+
+	"k8s.io/client-go/informers"
+	v1 "k8s.io/client-go/listers/core/v1"
```
kosmoz, line 23: maybe rename to `corev1`

internal/web/configurationvalues.go
```diff
@@ -72,10 +92,257 @@ func extractObjectKeyValueSource(r *http.Request, valueName string) *v1alpha1.Ob
...
+// the refKind is Secret, the datalist contains the secrets of the given namespace; if no namespace is given or an
+// error occurs, an empty datalist is returned
+func (s *server) namesDatalist(w http.ResponseWriter, r *http.Request) {
+	valueName := mux.Vars(r)["valueName"]
+	refKind := r.FormValue(refKindKey)
+	id := r.FormValue("id")
+	nsKey := formKey(valueName, namespaceKey)
+	namespace := r.Form.Get(nsKey)
+	var options []string
+	if err := r.ParseForm(); err != nil {
```
kosmoz, line 173: I don't understand why this is necessary here. Shouldn't the form be parsed before it is accessed using `FormValue`? Or am I missing something?

internal/web/server.go
```diff
@@ -100,6 +104,9 @@ type server struct {
 	packageController     cache.Controller
 	packageInfoStore      cache.Store
 	packageInfoController cache.Controller
+	namespaceLister       *v1.NamespaceLister
+	configMapLister       *v1.ConfigMapLister
+	secretLister          *v1.SecretLister
```
kosmoz, line 109: I'm not sure about best practices here, but since this feature is based on shared informers, wouldn't it be easier to just store the factory in the server object and then create listers on demand?

kosmoz (commented):
Nice work, although I have some questions below.

Also, I think we should start considering a different approach for the whole server object because I'm not sure how long the current approach is going to scale :sweat_smile: Maybe we can learn something from looking at other projects in the space. But that is a task for another day.

internal/web/server.go
christophenne: The problem with this approach is that the factory has to be started at some point, and inside the `Start()` method it needs to know the registered informers. When creating a lister with this factory after it has been started, it never returns any results because the informer is never started. 
We could initialize the three relevant listers inside `initWhenBootstrapped` before calling `Start()`, like this:
```
factory := informers.NewSharedInformerFactory(server.k8sClient, 0)
c := make(chan struct{})
factory.Core().V1().ConfigMaps().Lister()
factory.Start(c)
server.informerFactory = &factory
```
This way it would work by getting the listers "on demand" (which they wouldn't really be, because they are stored in the factory anyway), but I don't like it because if you need another lister at some other point in the future, you need to be aware of initializing it here first. Plus it doesn't seem right to request a lister here and then do nothing with it? But maybe I'm missing something.

kosmoz: Ohh, I didn't catch that! I thould it would actually start the informers on demand. 

In that case, I think the current approach is fine.

kosmoz (approved):

PR #696, opened 2024-05-25
--output support for glasskube install 

### LINKED ISSUE #656: Add `--output` support for `glasskube install`
See #629 for details.

Since `install` modifies a resource, the output of the command should be the created package.

Aryakoste:
Closes #656 

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -23,6 +24,7 @@ import (
 	"github.com/glasskube/glasskube/pkg/install"
 	"github.com/glasskube/glasskube/pkg/statuswriter"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
 )
 
 var installCmdOptions = struct {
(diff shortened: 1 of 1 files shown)


cmd/glasskube/cmd/install.go
```diff
@@ -200,9 +204,36 @@ var installCmd = &cobra.Command{
...
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "❗ Error: %v\n", err)
+				cliutils.ExitWithError()
+			}
+			fmt.Println(output)
+		}
 	},
 }
 
+func formatOutput(pkg *v1alpha1.Package, format OutputFormat) (string, error) {
```
christophenne, line 218: could you please make sure here to add the `GroupVersionKind` to the `pkg` if missing? Like here: https://github.com/glasskube/glasskube/pull/670/commits/18e921c8911ef438cac964d9b07254a3c8a3d68d

christophenne (changes requested):
Thanks, looking good!

One more thing I noticed when testing: There is a line printed at the top of the output to `stdout`: "✅ cyclops is now installed in dev." – it's not from your changes, but could you please change this information to be printed to `stderr` instead? That would be nice, thanks.

christophenne (approved):

PR #704, opened 2024-05-27
Christoph/glasskube cloud

christophenne:
## ✅ Checks


website/docusaurus.config.ts
```diff
@@ -1,8 +1,8 @@
-import {themes as prismThemes} from 'prism-react-renderer';
-import type {Config} from '@docusaurus/types';
-import type {Options as IdealImageOptions} from '@docusaurus/plugin-ideal-image';
+import { themes as prismThemes } from 'prism-react-renderer';
```
pmig, line 1: In most of the pages we don't use white spaces here, maybe we need to edit the .editorconfig or something in order to make sure autoformat doesn't add spaces here.

pmig (changes requested):
Once the cloud info is dismissed it doesn't come back. I would suggest to move our nav entries:

1. create a three column footer
2. move the current context to the footer
3. add a Glasskube Cloud link to the original current context position
4.  add a nother Glasskube Cloud link to the footer

For now I would just make the alert appear again. While we are working on a wording for the website we will probably change the link / target.

website/src/theme/cloud-banner.js
```diff
@@ -0,0 +1,15 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+export function onRouteDidUpdate({ location, previousLocation }) {
+  // Don't execute if we are still on the same page; the lifecycle may be fired
+  // because the hash changes (e.g. when navigating between headings)
+  if (location.pathname !== previousLocation?.pathname) {
+    if (ExecutionEnvironment.canUseDOM) {
+      const id = window.posthog?.get_distinct_id();
```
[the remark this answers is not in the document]
christophenne: You need to set `enableInDevelopment: true` in the `docusaurus.config.ts`

website/docusaurus.config.ts
kosmoz: adding this to the package.json would enforce this but I kind of prefer the spaces tbh :smile: 
```
  "prettier": {
    "bracketSpacing": false
  },
```

kosmoz (approved):

website/docusaurus.config.ts
christophenne: I wouldn't say most, there are quite a few with spaces too (probably the ones I edited?).  I have no preference, but I would definitely make it consistent.

kosmoz (approved):

pmig (approved):

PR #714, opened 2024-05-28
feat(ui): show reaction count on package detail page (#207)

### LINKED ISSUE #691: Discussion page is broken in multi repo environment
**Describe the bug**
Discussion page is broken for packages that do not appear in the public glasskube repo.

**Expected behavior**
The discussion page should be disabled and not linked to for such packages.

christophenne:
Ref #207 

Closes #691 

## ✅ Checks

### DIFF
--- a/api/v1alpha1/packagerepository_types.go
+++ b/api/v1alpha1/packagerepository_types.go
@@ -17,6 +17,9 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strings"
+
+	"github.com/glasskube/glasskube/internal/constants"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -91,6 +94,10 @@ func (repo *PackageRepository) SetDefaultRepositoryBool(value bool) {
 	} else {
 		delete(repo.Annotations, defaultRepositoryAnnotation)
 	}
+}
+
+func (repo *PackageRepository) IsGlasskubeRepo() bool {
+	return strings.HasPrefix(repo.Spec.Url, constants.DefaultRepoUrl)
 }
 
 //+kubebuilder:object:root=true
--- a/internal/giscus/client.go
+++ b/internal/giscus/client.go
@@ -1,0 +1,111 @@
+package giscus
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"net/url"
+	"sync"
+	"time"
+
+	"github.com/glasskube/glasskube/internal/httperror"
+)
+
+var client *GiscusClient
+
+type giscusConfig struct {
+	Repo       string
+	RepoId     string
+	Category   string
+	CategoryId string
+}
+
+type GiscusClient struct {
+	Config      *giscusConfig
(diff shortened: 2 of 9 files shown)


internal/giscus/client.go
```diff
@@ -0,0 +1,113 @@
...
+		values.Add("strict", "false")
+		values.Add("first", "1")
+		values.Add("number", "0")
+		u.RawQuery = values.Encode()
+
+		request, err := http.NewRequest(http.MethodGet, u.String(), nil)
+		if err != nil {
+			return nil, err
+		}
+		resp, err := http.DefaultClient.Do(request)
```
kosmoz, line 97: This can be replaced with a single call to `http.Get`

api/v1alpha1/packagerepository_types.go
```diff
@@ -93,6 +96,10 @@ func (repo *PackageRepository) SetDefaultRepositoryBool(value bool) {
 	}
 }
 
+func (repo *PackageRepository) IsGlasskubeRepo() bool {
+	return strings.Contains(repo.Spec.Url, constants.DefaultRepoUrl)
```
kosmoz, line 100: Maybe change to `strings.HasPrefix`?

kosmoz (commented):
I really like the looks of the new discussion link, and the new Giscus client looks neat as well! 

I think the "back" button on the discussion page looks a little out of place though. I think above the header or below the header might be better? 
image

And maybe another one below the comments?

christophenne:
I think the back button in these positions would always look a bit out of place, because the icons are always different, and for some it just looks a bit off. I now opted for removing the back button, but at least making icon and the title a link to the package detail page.

internal/giscus/client.go
```diff
@@ -0,0 +1,109 @@
...
+		values := u.Query()
+		values.Add("repo", c.Config.Repo)
+		values.Add("term", pkgName)
+		values.Add("category", c.Config.Category)
+		values.Add("strict", "false")
+		values.Add("first", "1")
+		values.Add("number", "0")
+		u.RawQuery = values.Encode()
+
+		resp, err := http.Get(u.String())
```
kosmoz, line 93: can be wrapped with `httperror.CheckResponse` to catch HTTP errors: 
```
httperror.CheckResponse(http.Get(u.String()))
```

kosmoz (commented):
Code looks good (I found another small improvement).

I noticed that the button seems to be flickering a little when changing between sites. It's not a huge issue IMO, so we can merge without addressing this, but if you know a way to prevent this it would be cool

christophenne:
> Code looks good (I found another small improvement).
> 
> I noticed that the button seems to be flickering a little when changing between sites. It's not a huge issue IMO, so we can merge without addressing this, but if you know a way to prevent this it would be cool

I tried to keep the flickering at a minimum by only loading the badge dynamically. Even if we only replace the content of it and have a placeholder while loading, it still feels like flickering unfortunately, but a little less I hope. Otherwise we could also check whether a number has already been cached for the current package, and then render the number directly instead of dynamically, but I don't know if that's worth it.

kosmoz (approved):
OK, as I already said, I think it's fine for now.

PR #723, opened 2024-05-31
feat(cli): Now bootstrap command will ask for user confirmation. (#719)

### LINKED ISSUE #719: Add confirmation to `bootstrap` command
The bootstrap command starts it's work right away, which can be dangerous if a user has the wrong kubeconfig context selected.

Before doing anything, `bootstrap` should ask the user for confirmation, the prompt should include the current context name. For example:

    Glasskube will be installed in context [name-of-current-context].
    Continue? (Y/n)

There should also be a `--yes` flag, that skips this step.

If the bootstrap process is triggered via the UI, the confirmation prompt MUST NOT be shown!

bishalbera:
Closes #719

## 📑 Description

Now the bootstrap command will ask for user confirmation if they want to use the current context and also they can omit the process with `--yes` flag

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/glasskube/glasskube/internal/clicontext"
 	"github.com/glasskube/glasskube/internal/clientutils"
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/config"
@@ -20,6 +21,7 @@ type bootstrapOptions struct {
 	disableTelemetry        bool
 	force                   bool
 	createDefaultRepository bool
+	yes                     bool
 }
 
 var bootstrapCmdOptions = bootstrapOptions{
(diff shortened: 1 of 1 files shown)


bishalbera:
@christophenne Hey I think the check failed because i used one capital letter in the commit message. Is it true? should I change the commit message and force push it .

christophenne:
> @christophenne Hey I think the check failed because i used one capital letter in the commit message. Is it true? should I change the commit message and force push it .

Yes exactly, the first letter of the subject shouldn't be upper case.

bishalbera:
> > @christophenne Hey I think the check failed because i used one capital letter in the commit message. Is it true? should I change the commit message and force push it .
> 
> Yes exactly, the first letter of the subject shouldn't be upper case.

ok then I'm changing and force pushing it.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -37,6 +39,17 @@ var bootstrapCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg, _ := cliutils.RequireConfig(config.Kubeconfig)
 		client := bootstrap.NewBootstrapClient(cfg)
+		ctx := cmd.Context()
+
+		currentContext := clicontext.RawConfigFromContext(ctx).CurrentContext
+
+		if !bootstrapCmdOptions.yes {
+			confirmMessage := fmt.Sprintf("Glasskube will be installed in context [%s].\nContinue? ", currentContext)
+			if !cliutils.YesNoPrompt(confirmMessage, false) {
```
christophenne, line 48: I think it was meant to be default `true`, since the "Y" is upper case in the issue description. Also minor, but could you remove the brackets around the context name?

christophenne (commented):
Thanks @bishalbera, works as expected. When you have time, could you please have a look at the minor things I commented? Then we're ready to merge.

cmd/glasskube/cmd/bootstrap.go
bishalbera: Done please check now @christophenne

christophenne (approved):

PR #726, opened 2024-05-31
feat(ui): add support for advanced options (#716)

### LINKED ISSUE #716: Implement specific (multi repository) package upgrades / downgrades
As user I want to change the version of an installed package (and even install the package from a different repository)

Introduce a new repository dropdown for not yet installed and already installed packages and add a version select to already installed packages, which changes the package version.
I also want to be able to change the auto upgrade checkbox for already installed packages.

It is already possible to install / configure specific versions from the cli (`glasskube install keptn --version=v2.0.0-rc.1+1`)

**Additional context**
I already created a mockup of the app detail page:

Glasskube package version and repo switch

christophenne:
Closes #716

## ✅ Checks

### DIFF
--- a/internal/web/templates/components/pkg-config-advanced.html
+++ b/internal/web/templates/components/pkg-config-advanced.html
@@ -1,0 +1,88 @@
+{{ define "pkg-config-advanced" }}
+  <div class="mt-3 border-warning border border-1 rounded-1 p-2 bg-body-secondary" id="advanced-options">
+    <h2 class="text-reset">Advanced Options</h2>
+    <div class="alert alert-warning" role="alert">
+      <i class="bi bi-exclamation-triangle-fill text-warning me-1"></i>Changing the repository or downgrading the
+      package version is not recommended and should never be done except in dev/test environments. This could lead to
+      data loss and a broken cluster! <b>Use at your own risk!</b>
+    </div>
+
+    <form
+      hx-post="/packages/{{ .Manifest.Name }}/configure/advanced"
+      hx-target="#error-container"
+      hx-target-error="#error-container"
+      hx-select="div.alert"
+      hx-swap="innerHTML">
+      {{ if gt (len .Repositories) 1 }}
+        <label class="form-label" for="advanced-pkg-install-repository">Repository</label>
+        <div class="input-group mb-2">
+          <select
+            class="form-select"
+            id="advanced-pkg-install-repository"
+            name="repositoryName"
+            required
+            hx-select="#advanced-options"
+            hx-swap="outerHTML"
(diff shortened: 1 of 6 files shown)


pmig (changes requested):
I always like advanced options :nerd_face:

I encountered two problems:

- The Advanced options look off in dark mode:

image

- I was a able to create following bug:

1. Have multiple repositories, but no default repository
2. install a package that has no `repositoryName` set
3. Enable advanced option
4. Go to package detail and change version
5. Package detail package will break with following error:

> `An error occurred fetching repository : resource name may not be empty`

internal/web/templates/pages/settings.html
```diff
@@ -28,6 +28,38 @@ <h2 class="text-reset">Package Repositories</h2>
...
+              class="form-check-input"
+              type="checkbox"
+              role="switch"
+              name="enableAdvancedOptions"
+              id="enableAdvancedOptions" />
+            <label class="form-check-label mx-1 fw-semibold" for="enableAdvancedOptions">Advanced Options</label>
+          </div>
+          <i class="bi bi-exclamation-triangle-fill text-warning me-1"></i>
+          It is not recommended to use these except for testing / maintaining a Glasskube package. These should only
+          ever be used in dev/test environments, and never in production. <b>Use at your own risk!</b>
```
kosmoz, line 45: I think there is missing a short description about what the advanced options do.

Also, `<strong>` should be preferred over `b`.

kosmoz (commented):
Just a small comment in addition to what has already been said. 

Other than that, looking good.

pmig (approved):
The only small improvement I can think of is that if a package is installed from a specific repository / in a specific version that these values are the default values for the select inputs see:

image

internal/web/templates/components/pkg-config-advanced.html
```diff
@@ -0,0 +1,88 @@
...
+            id="advanced-pkg-install-repository"
+            name="repositoryName"
+            required
+            hx-select="#advanced-options"
+            hx-swap="outerHTML"
+            hx-target="#advanced-options"
+            hx-get="/packages/{{ .Manifest.Name }}/configure/advanced"
+            hx-include="#advanced-pkg-install-version">
+            {{ range .Repositories }}
+              <option value="{{ .Name }}" {{ if eq .Name $.RepositoryName }}selected{{ end }}>
```
kosmoz, line 30: I noticed that the current repository is not selected initially, even though it should be, according to this.

kosmoz (changes requested):
I noticed a minor bug, but otherwise looking good.

PR #727, opened 2024-05-31
feat(cli): add `--dry-run` support for `glasskube install` to simulate package installation

### LINKED ISSUE #659: Add `--dry-run` support for `glasskube install`
See #629 for details.

The `--dry-run` flag can be used by users, who want to simulate installing a package, but don't actually want to install it.

Aryakoste:
Closes #659 

## ✅ Checks


cmd/glasskube/cmd/install.go
```diff
@@ -279,6 +280,8 @@ func init() {
 	installCmd.PersistentFlags().StringVar(&installCmdOptions.Repository, "repository", installCmdOptions.Repository,
 		"specify the name of the package repository to install this package from")
 	installCmd.PersistentFlags().BoolVar(&installCmdOptions.NoWait, "no-wait", false, "perform non-blocking install")
+	installCmd.PersistentFlags().BoolVar(&installCmdOptions.DryRun, "dry-run", false,
```
kosmoz, line 283: You can also mark `dry-run` and `no-wait` as mutually exclusive, since `dry-run` implies `no-wait`.

pkg/install/install.go
```diff
@@ -35,6 +37,14 @@ func (obj *installer) InstallBlocking(ctx context.Context, pkg *v1alpha1.Package
 	if err != nil {
 		return nil, err
 	}
+	if obj.dryRun {
+		time.Sleep(6 * time.Second)
```
kosmoz, line 41: I think waiting is not necessary for this feature, since `dry-run` is intended to be mostly used in automation.

pkg/install/install.go
```diff
@@ -50,6 +60,11 @@ func (obj *installer) install(
 	ctx context.Context,
 	pkg *v1alpha1.Package,
 ) (*v1alpha1.Package, error) {
+	if obj.dryRun {
+		obj.status.SetStatus(fmt.Sprintf("Dry run: Simulating installation of %v...", pkg.Name))
+		time.Sleep(6 * time.Second)
```
kosmoz, line 65: Again, please remove this `Sleep`.

Additionally, I think it would be great if we could take advantage of the kubernetes APIs dry-run feature, which can be enabled by setting the property in the create options of the request. However, our handwritten `PackageClient` does not support these `CreateOptions` yet. Would you be interested in adding this feature to the client and then adapting your PR to use it?

You may take the official kubernetes client as a reference: https://github.com/kubernetes/client-go/blob/52e5651101edcb2ecd8463ffdc281053cf6e63d4/kubernetes/typed/apps/v1/deployment.go#L123-L134

kosmoz (changes requested):

pkg/install/install.go
Aryakoste: Yeah I can do it. Will commit the changes. Once done.

kosmoz: Awesome, you rock :guitar:

Aryakoste:
I have added CreateOptions. But the package installtion gets stuck on Installing. I think awaitInstall isnt returning status as Ready. @kosmoz

kosmoz:
> I have added CreateOptions. But the package installtion gets stuck on Installing. I think awaitInstall isnt returning status as Ready. @kosmoz

Looks great! I think we have to just skip the await step in case dry-run is requested, as the package will never become ready.

Aryakoste:
if dry run set, I return status as nil is this ok or should i return package status as Ready like this.
return &client.PackageStatus{
            Status:  "Ready",
            Reason:  "DryRun",
            Message: "This is a simulated dry run.",
        }, nil

Aryakoste:
@kosmoz

pkg/install/install.go
```diff
@@ -28,30 +29,37 @@ func (obj *installer) WithStatusWriter(sw statuswriter.StatusWriter) *installer
...
 // the package has either status Ready or Failed.
-func (obj *installer) InstallBlocking(ctx context.Context, pkg *v1alpha1.Package) (*client.PackageStatus, error) {
+func (obj *installer) InstallBlocking(ctx context.Context, pkg *v1alpha1.Package,
+	opts metav1.CreateOptions) (*client.PackageStatus, error) {
 	obj.status.Start()
 	defer obj.status.Stop()
-	pkg, err := obj.install(ctx, pkg)
+	if isDryRun(opts) {
+		_, err := obj.install(ctx, pkg, opts)
+		return nil, err
```
kosmoz, line 38: This would lead to an error if err is nil. I think it would be best not to change the contract of this function, but you can also change the relevant behavior in `cmd/glasskube/cmd/install.go` so that `status == nil` is only considered an error if dry run has not been requested.

kosmoz (changes requested):

kosmoz:
> if dry run set, I return status as nil is this ok or should i return package status as Ready like this. return &client.PackageStatus{ Status: "Ready", Reason: "DryRun", Message: "This is a simulated dry run.", }, nil

Sorry, I didn't see your comment until after I submitted my review. Yes, that would be fine too.

Aryakoste:
For setting the flags mutually exclusive I am adding the code below in init.

installCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error { 		if installCmdOptions.NoWait && installCmdOptions.DryRun { 			return errors.New("the --dry-run and --no-wait flags are mutually exclusive") 		} 		return nil 	}

But it gives segmentation violation error.

Aryakoste:
@kosmoz

Aryakoste:
@kosmoz Changes done.

kosmoz (approved):

christophenne (approved):

PR #739, opened 2024-06-03
fix(package-operator): mark dependency as "waitingFor" if not found

kosmoz:
## 📑 Description
This PR fixes the following error:
```
> go run -race ./cmd/glasskube install keptn --yes --non-interactive --version v2.0.0+1
Summary:
 * The following packages will be installed in your cluster (minikube):
    1. keptn (version v2.0.0+1)
    2. cert-manager (version v1.14.5+1)
 * Automatic updates will be not enabled
⣻ Installing keptn...  [0s] ❌ keptn installation has status Failed, reason: InstallationFailed
Message: failed to get required package cert-manager: Package.packages.glasskube.dev "cert-manager" not found
```

The status is set to failed incorrectly, because the package operator has the responsibility to create the dependency. 

## ✅ Checks

## ℹ Additional Information

This error occurs because the package-controller uses the controller-runtime client which uses a cache for read operations. This means that a Get operation shortly after a resource has been created might fail, because the cache has not been updated yet.

### DIFF
--- a/internal/controller/package_controller.go
+++ b/internal/controller/package_controller.go
@@ -344,10 +344,14 @@ func (r *PackageReconcilationContext) ensureDependencies(ctx context.Context) bo
 			Namespace: r.pkg.Namespace,
 			Name:      dep.Name,
 		}, &requiredPkg); err != nil {
-			message := fmt.Sprintf("failed to get required package %v: %v", dep.Name, err)
-			r.setShouldUpdate(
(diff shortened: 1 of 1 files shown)


christophenne (commented):
Just for my understanding: even though the previous `CreateOrUpdate` call succeeds, the kube api server does not guarantee that it has actually created the resource?

kosmoz:
> Just for my understanding: even though the previous `CreateOrUpdate` call succeeds, the kube api server does not guarantee that it has actually created the resource?

Not quite, the client-side cache of the controller-runtime client does not have this guarantee.

christophenne (approved):

PR #740, opened 2024-06-03
feat(website): add badge to indicate default repository (#733)

### LINKED ISSUE #733: Indicate default repository via UI
Currently it is not possible to determine via the ui, if a repository is configured as default repository:

image

Add an indication to the UI that a repository is configured as default repository.

Currently it is necessary to describe the custom resource.

**Additional context**
Whether a given `PackageRepository` is marked as the default repository can be determined by checking  [`IsDefaultRepository()`](https://pkg.go.dev/github.com/glasskube/glasskube/api/v1alpha1#PackageRepository.IsDefaultRepository).

aminsaedi:
Closes #733 

## ✅ Checks

## ℹ Additional Information
I added a simple bootstrap badge to indicate default repository on the website

image

### DIFF
--- a/internal/web/templates/pages/settings.html
+++ b/internal/web/templates/pages/settings.html
@@ -19,7 +19,7 @@
                     {{ end }}
                   </div>
                   <div class="ms-2 d-flex flex-column">
-                      <span class="fw-semibold">
+                    <span class="fw-semibold">
                       {{ .Name }}
                       {{ if IsRepoStatusReady . }}
                         <span class="badge bg-primary">Default</span>
--- a/internal/web/templates/pages/settings.html
+++ b/internal/web/templates/pages/settings.html
@@ -21,7 +21,7 @@
                   <div class="ms-2 d-flex flex-column">
                     <span class="fw-semibold">
                       {{ .Name }}
-                      {{ if IsRepoStatusReady . }}
+                      {{ if .IsDefaultRepository }}
                         <span class="badge bg-primary">Default</span>
                       {{ end }}
                     </span>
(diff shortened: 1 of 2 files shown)


internal/web/templates/pages/settings.html
```diff
@@ -19,7 +19,12 @@ <h2 class="text-reset">Package Repositories</h2>
                     {{ end }}
                   </div>
                   <div class="ms-2 d-flex flex-column">
-                    <span class="fw-semibold">{{ .Name }}</span>
+                    <span class="fw-semibold">
+                      {{ .Name }}
+                      {{ if IsRepoStatusReady . }}
```
christophenne, line 24: I think you intended to call your `IsDefaultRepository` function here instead of the `IsRepoStatusReady`.

This can be simplified, because the `PackageRepository` type already has a `IsDefaultRepository` function (see `api/v1alpha1/packagerepository_types.go`, and you can call such functions directly from the template. So you can remove the `IsDefaultRepository` function from the `templateFuncs`, and simply call `.IsDefaultRepository` in the template.

christophenne (changes requested):
Hi @aminsaedi, thanks for taking this up! I made a comment about your implementation, could you please incorporate these changes? Thanks!

pmig (changes requested):
Please also rename the commit messages to `feat(ui): ...` as `feat(website): ...` should only be used for changes in den `website/` folder which will be deployed to https://glasskube.dev/ :-)

internal/web/templates/pages/settings.html
aminsaedi: Thanks for your suggestion, It's now updated!

christophenne (approved):

christophenne:
> I used git rebase interactive mode to change the commit messages, then I had to do a force push to update the history in the remote, please let me know if there is any other way to do it

Yes this was correct, but we prefer to squash all commits before merging anyway – we will do that on the Github UI for this PR.

pmig (approved):

Issue #754, opened 2024-06-05
PR Workflow Update: Migrate from Rebase-Merge to Squash-Merge


kosmoz:
Squash merging provides several benefits over rebasing:
- always a single changelog entry for a PR
- no need for contributors to continuously rebase their branch and fiddle with commit message format
- more control over commit message in `main` branch (and thus changelog entries) for the core team

In order to adopt this workflow, the following changes must be made:
- [ ] Disable all merge options other than "Squash & Merge" in the repository settings (must be performed by a repository admin)
- [ ] Add a PR check that verifies that the title conforms to the spec. There are two options for this, an app and a workflow action:
      - https://github.com/Ezard/semantic-prs (used by renovate)
      - https://github.com/amannn/action-semantic-pull-request
- [ ] Remove the commitlint workflow

kosmoz:
Personally, I disagree. Considering the example from your comment, I think the acceptable strategy would be creating multiple PRs in parallel

PR #775, opened 2024-06-07
ci: prepare for semantic PR workflow

kosmoz:
Ref #754  

## 📑 Description
This PR includes several changes to the project's CI workflows. In the new concept, a merge queue will be used to merge PRs into main, so we don't always have to update branches before merging. This requires that some checks are required and those checks must include the `merge_group` trigger. Because required checks must also run on the pull request _before_ merging, we use different jobs with the same name to execute similar tasks for pull requests and merge groups: 
- `build` always builds the project, but it builds the website for changes in the website directory and the go project for all other changes. `build` is already configured as a required check.
-  `semantic-message` checks whether the PR title conforms with the conventional commit message format during PR phase and checks if the commit message conforms with the same format for merge groups. `semantic-message` is a new job name and is not yet configured as a required check.

## ✅ Checks

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,6 +7,11 @@ on:
   pull_request:
     paths-ignore:
       - website/**
+  merge_group:
+    types:
+      - checks_requested
+    paths-ignore:
+      - website/**
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


pmig (approved):
LGTM

christophenne (approved):

PR #779, opened 2024-06-08
feat(cli): add `--output` flag for `glasskube bootstrap`

### LINKED ISSUE #724: Add `--output` support for `glasskube bootstrap`

hanshal101:
Closes #724 
Fixes #724 

## 📑 Description
This PR adds the feature for receiving JSON/YAML output for bootstrap command. This gives an output of all the Kubernetes Resources that are changed. 

## ✅ Checks

## ℹ Working
1. Give a bootstrap command with the flags `-o` add YAML or JSON to it
2. Now it will first bootstrap the cluster as it will normally do (unless an error occurs)
3. After the bootstrap is completed successfully it will show all the changed Resources in the format you gave in.
something like this:

```
## Installing GLASSKUBE ##
≡ƒºè The next generation Package Manager for Kubernetes ≡ƒôª
* Fetching Glasskube manifest from https://github.com/glasskube/glasskube/releases/download/v0.8.0/manifest-aio.yaml
* Validating existing installation
* Applying Glasskube manifests
* Telemetry is enabled for this cluster ΓÇô Run "glasskube telemetry status" for more info.
* Glasskube successfully installed! (took 22s)
apiVersion: v1
kind: Namespace
metadata:
  name: flux-system
---
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    packages.glasskube.dev/telemetry-enabled: "true"
    packages.glasskube.dev/telemetry-id: b753e233-32ac-47eb-a25b-299b74094587
  labels:
    app.kubernetes.io/component: manager
    app.kubernetes.io/created-by: glasskube
    app.kubernetes.io/instance: system
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: namespace
    app.kubernetes.io/part-of: glasskube
    control-plane: controller-manager
  name: glasskube-system
---
apiVersion: v1
kind: ResourceQuota
metadata:
  name: critical-pods
  namespace: flux-system
spec:
  hard:
    pods: "1000"
  scopeSelector:
    matchExpressions:
    - operator: In
      scopeName: PriorityClass
      values:
      - system-node-critical
      - system-cluster-critical
---

// Rest all Resources/Objects
```

### DIFF
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -135,7 +135,7 @@ func convertAndPrintManifests(
 		}
 	case OutputFormatYAML:
 		for i, obj := range objs {
-			yamlData, err := yaml.Marshal(obj)
+			yamlData, err := yaml.Marshal(&obj)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "error marshaling data to YAML: %v\n", err)
 				cliutils.ExitWithError()
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -90,7 +90,7 @@ var bootstrapCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "\nAn error occurred during bootstrap:\n%v\n", err)
 			cliutils.ExitWithError()
 		}
-		if err := bootstrapCmdOptions.printBootsrap(
+		if err := printBootsrap(
 			manifests,
 			bootstrapCmdOptions.Output,
 		); err != nil {
@@ -111,8 +111,8 @@ func (o bootstrapOptions) asBootstrapOptions() bootstrap.BootstrapOptions {
 	}
 }
 
-func (o bootstrapOptions) printBootsrap(manifests []unstructured.Unstructured, output OutputFormat) error {
-	if o.Output != "" {
+func printBootsrap(manifests []unstructured.Unstructured, output OutputFormat) error {
+	if output != "" {
 		if err := convertAndPrintManifests(manifests, output); err != nil {
 			return err
 		}
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -12,6 +12,9 @@ import (
 	"github.com/glasskube/glasskube/internal/util"
 	"github.com/glasskube/glasskube/pkg/bootstrap"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	js "k8s.io/apimachinery/pkg/runtime/serializer/json"
 )
 
 type bootstrapOptions struct {
@@ -79,14 +82,17 @@ var bootstrapCmd = &cobra.Command{
 			}
 
 		}
-		if err := client.Bootstrap(
+		manifests, err := client.Bootstrap(
 			cmd.Context(),
 			bootstrapCmdOptions.asBootstrapOptions(),
-			bootstrapCmdOptions.Output.String(),
-		); err != nil {
+		)
+		if err != nil {
 			fmt.Fprintf(os.Stderr, "\nAn error occurred during bootstrap:\n%v\n", err)
 			cliutils.ExitWithError()
-
+		}
+		if err := bootstrapCmdOptions.printBootsrap(manifests, bootstrapCmdOptions.Output); err != nil {
+			fmt.Fprintf(os.Stderr, "\n Error occured in printing : %v\n", err)
+			cliutils.ExitWithError()
 		}
 	},
 }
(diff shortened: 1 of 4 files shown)


hanshal101:
I'm not really convinced with the output, I think on hitting the bootstrap command with the output flags it should directly print the resources and objects. But can't find any other way to solve this.

pkg/bootstrap/bootstrap.go
```diff
@@ -140,6 +144,44 @@ func (c *BootstrapClient) Bootstrap(ctx context.Context, options BootstrapOption
...
+		opt = false
+		div = ","
+	case "yaml":
+		opt = true
+		div = "---"
+	}
+	serializer := js.NewSerializerWithOptions(
+		js.DefaultMetaFactory, scheme, scheme,
+		js.SerializerOptions{Yaml: opt, Pretty: true, Strict: true},
+	)
```
kosmoz, line 173: We don't usually use the `serializer` package, instead we use the `json` an `yaml` packages directly. Please use those instead. Also, please make sure that the format is consistent with the other commands: JSON output should be printed as a JSON array, while YAML output should be printed in YAML multi document mode (like you already do).

pkg/bootstrap/bootstrap.go
```diff
@@ -140,6 +144,44 @@ func (c *BootstrapClient) Bootstrap(ctx context.Context, options BootstrapOption
...
+
+	for _, obj := range objs {
+		runtimeObj := &unstructured.Unstructured{}
+		obj.DeepCopyInto(runtimeObj)
```
kosmoz, line 177: I don't understand why this is necessary. Can this be removed?

kosmoz (changes requested):
Hi @hanshal101, please review my comments below and let me know if you have any questions

hanshal101:
Hey @kosmoz can you help me with this error in build.

kosmoz:
> Hey @kosmoz can you help me with this error in build.

Here is the error message: 
```
 Error: cmd/glasskube/cmd/bootstrap.go:97:39: `occured` is a misspelling of `occurred` (misspell)
			fmt.Fprintf(os.Stderr, "\nAn error occured in printing : %v\n", err)
```

You have to change "occured" to "occurred"

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -97,12 +111,55 @@ func (o bootstrapOptions) asBootstrapOptions() bootstrap.BootstrapOptions {
 	}
 }
 
+func (o bootstrapOptions) printBootsrap(manifests []unstructured.Unstructured, output OutputFormat) error {
```
kosmoz, line 114: the output format is used twice here but taken from different places.

I think the `bootstrapOptions` receiver should be removed here and then the function should check `if output != ""` instead.

kosmoz (changes requested):
Code looks good now, but I found two bugs (see example below):
1. This was not an issue until now, but some parts of `pkg/bootstrap` write to `os.Stdout`, where they should write to `os.Stderr`. 
2. The output is a little off, particularly, the `Object` key should be missing. I believe the reason why this is happening is that there is only a `MarshalJSON` function for `*Unstructured`, not for `Unstructured`, so it is not called here. This can be prevented by calling `yaml.Marshal` with `&obj` instead of `obj`. Interestingly, it works well enough for JSON.

Example output:
```
## Installing GLASSKUBE ##
🧊 The next generation Package Manager for Kubernetes 📦
* Fetching Glasskube manifest from https://github.com/glasskube/glasskube/releases/download/v0.8.0/manifest-aio.yaml
* Validating existing installation
* Applying Glasskube manifests
* Glasskube successfully installed! (took 57s)
Object:
  apiVersion: v1
  kind: Namespace
  metadata:
    name: flux-system
---

...
```

hanshal101:
> 1. Stdout

Are you sure because I can't find any `os.Stdout` in `glasskube\pkg\bootstrap\bootstrap.go`
I think you meant in the `cmd/bootstrap.go`

kosmoz:
> > 1. Stdout
> 
> Are you sure because I can't find any `os.Stdout` in `glasskube\pkg\bootstrap\bootstrap.go` I think you meant in the `cmd/bootstrap.go`

Yes, I'm sure. `fmt.Print*` functions write to stdout. They should be replaced with `fmt.Fprint*` and `os.Stderr`.

hanshal101:
Do i have to change it over here:
```
func statusMessage(input string, success bool) {
	if success {
		green := color.New(color.FgGreen).SprintFunc()
		fmt.Println(green("* " + input))
	} else {
		red := color.New(color.FgRed).SprintFunc()
		fmt.Println(red("* " + input))
	}
}
```

kosmoz:
> Do i have to change it over here:
> 
> ```
> func statusMessage(input string, success bool) {
> 	if success {
> 		green := color.New(color.FgGreen).SprintFunc()
> 		fmt.Println(green("* " + input))
> 	} else {
> 		red := color.New(color.FgRed).SprintFunc()
> 		fmt.Println(red("* " + input))
> 	}
> }
> ```

Yes, and possibly one or two other places in the same file

hanshal101:
I have made the necessary changes; if they are correct, then I'll rebase

kosmoz (approved):

PR #808, opened 2024-06-13
feat(cli): add `repo update` command

### LINKED ISSUE #682: Support repo updates with `glasskube repo update [repoName]`
Currently, there is no way to update a `PackageRepository` custom resource via our CLI.

There should be a `repo`-subcommand `delete`, which can be used like this:

`glasskube repo update [repositoryName]`

It should support the same flags as the `repo add` command, and additionally the `--url` flag to make changing the url possible. 

**Additional context**
Ref #195

doncicuto:
Closes #682 

## 📑 Description
This PR adds the update option to `glasskube repo` command

## ✅ Checks
- [X] I have updated the documentation as required

## ℹ Additional Information

This PR has the following important changes:
- A none auth type has been added to both the `repo add` and `repo update`command. None auth type is mainly used to specify that we no longer require authentication when updating a repo
- If --default is used, the current default repo gets the default annotation removed, and the repo being updated gets the default annotation added. If an error occurs when this operations are perfomed a rollback is performed
- A fix has been added to Normalize that checks if wrong flags have been set that don't apply for a specific auth type

Here are some screenshots showing this option in action:

image

image


cmd/glasskube/cmd/repo_options.go
```diff
@@ -52,17 +60,72 @@ func (opts *repoOptions) BindToCmdFlags(cmd *cobra.Command) {
...
 		}
 	} else if len(opts.Token) > 0 {
-		if opts.Auth != repoAddNoAuth {
+		if opts.Auth != repoBearerAuth {
 			return fmt.Errorf("token only applies to bearer authentication (got %v)", opts.Auth)
 		} else {
-			opts.Auth = repoAddBearerAuth
+			opts.Auth = repoBearerAuth
+		}
+	}
```
kosmoz, line 81: This breaks a currently supported input:
```
glasskube repo add foo http://bar --username admin
```
This input implies that the auth type is "basic", since only basic auth uses the username value. 

In contrast, the following input was previously unsupported:
```
glasskube repo add foo http://bar --auth basic --username admin
```
Though I think it would make sense to support that as well.

You can change the conditions to `opts.Auth != repoNoAuth && opts.Auth != repo.BasicAuth` or something like that.

cmd/glasskube/cmd/repo_add.go
```diff
@@ -97,5 +53,6 @@ var repoAddCmd = &cobra.Command{
 }
 
 func init() {
-	repoAddCmdOptions.BindToCmdFlags(repoAddCmd)
+	update := false
```
kosmoz, line 56: should be inline

cmd/glasskube/cmd/repo_update.go
```diff
@@ -0,0 +1,86 @@
...
+			cliutils.ExitWithError()
+		}
+
+		fmt.Fprintf(os.Stderr, "✅ package repository %v updated\n", repoName)
+		cliutils.ExitSuccess()
+	},
+}
+
+func init() {
+	update := true
```
kosmoz, line 84: should be inline

internal/cliutils/packagerepository.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	for _, r := range repos.Items {
+		if r.IsDefaultRepository() {
+			defaultRepo = r
+			break
+		}
+	}
+
+	return &defaultRepo, nil
```
kosmoz, line 25: This would return an invalid value if there is no default repo. I think we should return an error in that case instead.

kosmoz (changes requested):

doncicuto:
Thanks @kosmoz for your review, I'll prepare the changes and update the PR

cmd/glasskube/cmd/repo_options.go
doncicuto: Done, thanks!

cmd/glasskube/cmd/repo_update.go
doncicuto: Done, thanks!

internal/cliutils/packagerepository.go
doncicuto: Done, thanks!

cmd/glasskube/cmd/repo_add.go
doncicuto: Done, thanks!

doncicuto:
Hi @kosmoz I've updated the PR hope the fixes look good to you, if you need any further changes or if I misunderstood any of the suggested changes let me know. Thanks!

kosmoz (approved):

Issue #803, opened 2024-06-12
Namespace-scoped packages


kosmoz:
In the current state, Glasskube packages can only be installed once per cluster. In some situations, this is an unnecessary restriction and limits some use-cases. 

It should be possible for the author of a package to specify a "scope", which can either be "Cluster" or "Namespaced" (default "Cluster"). Depending on a packages scope, the Glasskube system should create either a cluster or namespace scoped custom resource. The name of the cluster-scoped CRD is `ClusterPackage`, the name for the namespace-scoped CRD is `Package`. This is a breaking change, because we already currently the `Package` CRD name for the cluster-scoped resources, but we settled on this to stay consistent with common Kubernetes nomenclature (e.g. `Role`/`ClusterRole`). This change will be accompanied by an API version bump to `v1alpha2`. Existing installations can not be migrated automatically.

### Tasklist
- [ ] Refactor the codebase to support `ClusterPackage` and `Package`
- [ ] Update the dependency manager to allow `Package` validation
- [ ] Add UI for `Package` list, installation
- [ ] Update `install` command to allow creating `Package`
- [ ] Update `update` command to allow updating `Package`
- [ ] Update `configure` command to allow updating `Package`
- [ ] Update `describe` command to show status of all `Package` instances
- [ ] Update `list` command to show all `Package` instances
- [ ] Update `open` command to support `Package`
- [ ] Update `uninstall` command to allow deleting `Package`
- [ ] Update `autoupdate` commands to handle `Package`

### Limitations
* For now, dependencies must always be Cluster scoped, even for namespace scoped packages

PR #817, opened 2024-06-17
feat(ui): add support for namespace-scoped packages

christophenne:
Ref #803 

## ✅ Checks

### Important:
- [x] undo changes of https://github.com/glasskube/glasskube/pull/839


pkg/list/lister.go
```diff
@@ -39,55 +49,122 @@ func NewLister(ctx context.Context) *lister {
...
+	return !((options.OnlyInstalled && !item.Installed()) || (options.OnlyOutdated && !item.Outdated()))
+}
+
+type typeOptions int
+
+const (
+	// TODO is this how this works?
+	includeClusterPackages typeOptions = 1 << iota
+	includePackages
+	includeAll = 3
```
kosmoz, line 122: Regarding the TODO comment: Yes, this will set the first const to 1, the second to 2, the third to 4, etc.

I would change the last item to make it more explicit and give it the correct type :slightly_smiling_face: 
```go
includeAll = includeClusterPackages | includePackages
```

pkg/list/lister.go
```diff
@@ -101,27 +178,47 @@ func (l *lister) fetchRepoAndInstalled(ctx context.Context, options ListOptions)
 
 	wg.Wait()
 
-	compositeErr := multierr.Combine(repoErr, pkgErr, pkgInfoErr)
-	if pkgErr != nil || pkgInfoErr != nil {
+	compositeErr := multierr.Combine(repoErr, clPkgErr, pkgErr, pkgInfoErr)
+	if clPkgErr != nil || pkgErr != nil || pkgInfoErr != nil {
```
kosmoz, line 182: Already was like this, but I just realized that `repoErr` is not considered here. Should it be?

pkg/client/client_cached.go
```diff
@@ -35,6 +36,23 @@ func (c *cacheClientset) ClusterPackages() ClusterPackageInterface {
...
+	}
+	return &readWriteCacheClient[v1alpha1.Package, v1alpha1.PackageList]{
+		p,
+		readOnlyCacheClient[v1alpha1.Package, v1alpha1.PackageList]{
+			p,
+			c.packageStore,
+			func(items []v1alpha1.Package) v1alpha1.PackageList {
+				return v1alpha1.PackageList{Items: items}
+			},
+		},
```
kosmoz, line 52: This probably doesn't check the namespace, we should probably add this to the `readOnlyCacheClient`

pkg/list/lister.go
christophenne: I believe that was on purpose, because it could be that one of the repositories is not available, but the rest could still be displayed – or am I mixing this up with something else?

kosmoz: Yes I think you're right. Maybe there should be a comment explaining this :sweat_smile:

pkg/client/client_cached.go
christophenne: Good point, everything still worked but it always went into the fallback because the namespaced items were never found in the cache store. They are stored in this format: `namespace/name`. I tried to make the cache optionally namespace-aware, where the list-part is unused (we never `GetAll` on namespaced packages), but the `Get` case seems to work. Please have a look.

internal/repo/types/types.go
```diff
@@ -18,6 +18,7 @@ type PackageRepoIndexItem struct {
 	ShortDescription string `json:"shortDescription,omitempty"`
 	IconUrl          string `json:"iconUrl,omitempty"`
 	LatestVersion    string `json:"latestVersion,omitempty"`
+	Scope            string `json:"scope,omitempty"`
```
kosmoz, line 21: Could we use `v1alpha1.PackageScope` for this?

pkg/list/lister.go
```diff
@@ -101,27 +177,48 @@ func (l *lister) fetchRepoAndInstalled(ctx context.Context, options ListOptions)
...
-				packageInfoName := names.PackageInfoName(&pkg)
-				for k, packageInfo := range packageInfos.Items {
-					if packageInfo.Name == packageInfoName {
-						result[i].PackageInfo = &packageInfos.Items[k]
-						break
-					}
+		res := result{
+			IndexItem: &index.Packages[i],
+		}
+		if v1alpha1.PackageScope(indexPackage.Scope) == v1alpha1.ScopeCluster && typeOpts&includeClusterPackages != 0 {
```
kosmoz, line 193: Check for empty scope is missing. 

Since this is probably going to happen a lot, I think `PackageRepoIndexItem` should use `*PackageScope` and then we can define some functions like `(*PackageScope) IsCluster() bool` and `(*PackageScope) IsNamespaced() bool` on the type directly.

kosmoz (changes requested):

internal/web/server.go
```diff
@@ -18,6 +18,12 @@ import (
 	"sync"
 	"syscall"
 
+	types2 "k8s.io/apimachinery/pkg/types"
```
kosmoz, line 21: I think a better name conflict resolution would be changing the import name of "repo/types" to `repotypes` and leaving this one as `types` or maybe `apitypes`

internal/web/server.go
```diff
@@ -232,351 +262,416 @@ func (s *server) Start(ctx context.Context) error {
...
-	ctx := r.Context()
-	pkgName := r.FormValue("packageName")
-	var pkg v1alpha1.ClusterPackage
-	if err := s.pkgClient.ClusterPackages().Get(ctx, pkgName, &pkg); err != nil {
-		s.respondAlertAndLog(w, err, fmt.Sprintf("An error occurred fetching %v during uninstall", pkgName), "danger")
-		return
+func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrlpkg.Package) {
+	fwName := pkg.GetName()
+	if pkg.IsNamespaceScoped() {
+		fwName = cache.ObjectName{Namespace: pkg.GetNamespace(), Name: pkg.GetName()}.String()
```
kosmoz, line 464: This can be simplified, because the `ObjectName` does not put a `/` if the namespace is empty, resulting in the same outcome.

internal/web/server.go
```diff
@@ -638,59 +807,68 @@ func (s *server) advancedConfiguration(w http.ResponseWriter, r *http.Request) {
...
+		d.pkg.GetSpec().PackageInfo.Version = d.selectedVersion
+		if d.repositoryName != "" {
+			d.pkg.GetSpec().PackageInfo.RepositoryName = d.repositoryName
 		}
-		if err := s.pkgClient.ClusterPackages().Update(ctx, pkg); err != nil {
-			s.respondAlertAndLog(w, err,
-				fmt.Sprintf("An error occurred updating package %v to version %v in repo %v", pkgName, selectedVersion, repositoryName),
-				"danger")
-			return
+		if d.manifest.Scope.IsCluster() {
```
kosmoz, line 852: I think it would be better to use a type switch here:
```go
switch pkg := d.pkg.(type) {
  case *v1alpha1.Package:
    // ...
  case *v1alpha1.ClusterPackage:
    // ...
  default:
    panic("unexpected package type")
}
```

kosmoz (changes requested):

kosmoz (approved):

PR #819, opened 2024-06-18
feat(cli): add --dry-run to bootstrap command

### LINKED ISSUE #725: Add `--dry-run` support for `glasskube bootstrap`

hanshal101:
Closes #725
Fixes #725

## 📑 Description
In this PR I have added the `--dry-run` to the `glasskube bootstrap` command

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -29,7 +29,7 @@ type bootstrapOptions struct {
 	force                   bool
 	createDefaultRepository bool
 	yes                     bool
-	dryrun                  bool
+	dryRun                  bool
 	OutputOptions
 }
 
@@ -77,6 +77,12 @@ var bootstrapCmd = &cobra.Command{
 			}
 		}
 
+		if bootstrapCmdOptions.dryRun {
+			fmt.Fprintln(os.Stderr,
+				"🔎 Dry-run mode is enabled. "+
+					"Nothing will be changed in your cluster, but validations will still be run.")
+		}
+
 		verifyLegalUpdate(ctx, installedVersion, targetVersion)
 
 		manifests, err := client.Bootstrap(ctx, bootstrapCmdOptions.asBootstrapOptions())
@@ -102,7 +108,7 @@ func (o bootstrapOptions) asBootstrapOptions() bootstrap.BootstrapOptions {
 		DisableTelemetry:        o.disableTelemetry,
 		Force:                   o.force,
 		CreateDefaultRepository: o.createDefaultRepository,
-		DryRun:                  o.dryrun,
+		DryRun:                  o.dryRun,
 	}
 }
 
@@ -223,8 +229,8 @@ func init() {
 	bootstrapCmd.Flags().BoolVar(&bootstrapCmdOptions.createDefaultRepository, "create-default-repository",
 		bootstrapCmdOptions.createDefaultRepository,
 		"Toggle creation of the default glasskube package repository")
-	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOptions.dryrun, "dry-run", false,
-		"simulate the bootstrapping of Glasskube without actually bootstrapping it")
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOptions.dryRun, "dry-run", false,
+		"Do not make any changes but run all validations")
 	bootstrapCmd.Flags().BoolVar(&bootstrapCmdOptions.yes, "yes", false, "Skip confirmation prompt")
 	bootstrapCmd.MarkFlagsMutuallyExclusive("url", "type")
 	bootstrapCmd.MarkFlagsMutuallyExclusive("url", "latest")
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -84,22 +84,12 @@ var bootstrapCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "\nAn error occurred during bootstrap:\n%v\n", err)
 			cliutils.ExitWithError()
 		}
-		if bootstrapCmdOptions.dryrun {
-			if err := printBootstrap(
-				manifests,
-				bootstrapCmdOptions.Output,
-			); err != nil {
-				fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
-				cliutils.ExitWithError()
-			}
-		} else {
-			if err := printBootstrap(
-				manifests,
-				bootstrapCmdOptions.Output,
-			); err != nil {
-				fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
-				cliutils.ExitWithError()
-			}
+		if err := printBootstrap(
+			manifests,
+			bootstrapCmdOptions.Output,
+		); err != nil {
+			fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
+			cliutils.ExitWithError()
 		}
 	},
 }
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -85,7 +85,6 @@ var bootstrapCmd = &cobra.Command{
 			cliutils.ExitWithError()
 		}
 		if bootstrapCmdOptions.dryrun {
-			fmt.Println("Dry run completed successfully.")
 			if err := printBootsrap(
 				manifests,
 				bootstrapCmdOptions.Output,
@@ -236,7 +235,7 @@ func init() {
 		bootstrapCmdOptions.createDefaultRepository,
 		"Toggle creation of the default glasskube package repository")
 	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOptions.dryrun, "dry-run", false,
-		"simulate the installation of package without actually installing it")
+		"simulate the bootstrapping of Glasskube without actually bootstrapping it")
 	bootstrapCmd.Flags().BoolVar(&bootstrapCmdOptions.yes, "yes", false, "Skip confirmation prompt")
 	bootstrapCmd.MarkFlagsMutuallyExclusive("url", "type")
 	bootstrapCmd.MarkFlagsMutuallyExclusive("url", "latest")
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -92,6 +92,14 @@ var bootstrapCmd = &cobra.Command{
 				fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
 				cliutils.ExitWithError()
 			}
+		} else {
+			if err := printBootstrap(
+				manifests,
+				bootstrapCmdOptions.Output,
+			); err != nil {
+				fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
+				cliutils.ExitWithError()
+			}
 		}
 	},
 }
--- a/cmd/glasskube/cmd/bootstrap.go
+++ b/cmd/glasskube/cmd/bootstrap.go
@@ -29,6 +29,7 @@ type bootstrapOptions struct {
 	force                   bool
 	createDefaultRepository bool
 	yes                     bool
+	dryrun                  bool
 	OutputOptions
 }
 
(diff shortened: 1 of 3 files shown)


hanshal101:
Sorry for the late PR😅, will rebase it when everything looks good. I have taken some inspiration/ideas from the `--dry-run` in the install command.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -166,6 +180,8 @@ func init() {
 	bootstrapCmd.Flags().BoolVar(&bootstrapCmdOptions.createDefaultRepository, "create-default-repository",
 		bootstrapCmdOptions.createDefaultRepository,
 		"Toggle creation of the default glasskube package repository")
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOptions.dryrun, "dry-run", false,
+		"simulate the installation of package without actually installing it")
```
christophenne, line 184: The description is not accurate, please change it to "Simulate the installation of Glasskube without actually installing it"

pkg/bootstrap/bootstrap.go
```diff
@@ -132,6 +138,14 @@ func (c *BootstrapClient) Bootstrap(
 		manifests = append(manifests, defaultRepository())
 	}
 
+	if options.DryRun {
+		statusMessage("Dry-run mode enabled. The following resources would be applied:", true)
+		for _, manifest := range manifests {
```
christophenne, line 143: I think printing the manifest names here is not necessary, and also it would make the output yaml invalid this way.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -90,13 +91,25 @@ var bootstrapCmd = &cobra.Command{
...
 			cliutils.ExitWithError()
 		}
-		if err := printBootsrap(
-			manifests,
-			bootstrapCmdOptions.Output,
-		); err != nil {
-			fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
-			cliutils.ExitWithError()
+		if bootstrapCmdOptions.dryrun {
+			fmt.Println("Dry run completed successfully.")
```
christophenne, line 95: This prints the message "Dry run completed successfully." to `stdout`. If you run the bootstrap dryrun in combination with `--output yaml` (which is probably the main use case I think), this will put that exact string into the yaml, making it an invalid yaml. That's why such messages are always printed to `stderr` instead. 

Please test your solution like this: `glasskube bootstrap --dry-run --o yaml > output.yaml` and then check the content of the `output.yaml`. This must be valid yaml which could be used with `kubectl apply` for example (please test that this actually works).

christophenne (changes requested):
Thanks @hanshal101, please have a look at the comments.

cmd/glasskube/cmd/bootstrap.go
hanshal101: Ohk, so here the output comes something like this:
```
flux-system
glasskube-system
critical-pods
buckets.source.toolkit.fluxcd.io
gitrepositories.source.toolkit.fluxcd.io
helmcharts.source.toolkit.fluxcd.io
helmreleases.helm.toolkit.fluxcd.io
helmrepositories.source.toolkit.fluxcd.io
ocirepositories.source.toolkit.fluxcd.io
packageinfos.packages.glasskube.dev
packagerepositories.packages.glasskube.dev
packages.packages.glasskube.dev
helm-controller
source-controller
glasskube-controller-manager
glasskube-webhook-cert
glasskube-leader-election-role
glasskube-webhook-cert-role
crd-controller
flux-edit
flux-view
glasskube-metrics-reader
glasskube-packagerepository-editor-role
glasskube-packagerepository-viewer-role
glasskube-proxy-role
glasskube-webhook-cert-role
glasskube-leader-election-rolebinding
glasskube-webhook-cert-rolebinding
cluster-reconciler
crd-controller
glasskube-manager-rolebinding
glasskube-proxy-rolebinding
glasskube-webhook-cert-rolebinding
source-controller
glasskube-controller-manager-metrics-service
glasskube-webhook-service
helm-controller
source-controller
glasskube-controller-manager
glasskube-webhook-cert
glasskube-webhook-cert-init
allow-egress
allow-scraping
allow-webhooks
glasskube-validating-webhook-configuration
glasskube
Dry run completed successfully. The following manifests would be applied:
apiVersion: v1
kind: Namespace
metadata:
  name: flux-system
---
```

Here we don't require the texts above the `yaml`
The output should be something like this:
```
apiVersion: v1
kind: Namespace
metadata:
  name: flux-system

---
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    packages.glasskube.dev/telemetry-enabled: "true"
    packages.glasskube.dev/telemetry-id: b753e233-32ac-47eb-a25b-299b74094587
  labels:
    app.kubernetes.io/component: manager
    app.kubernetes.io/created-by: glasskube
    app.kubernetes.io/instance: system
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: namespace
    app.kubernetes.io/part-of: glasskube
    control-plane: controller-manager
  name: glasskube-system

---
```

christophenne: Yes, the output should only be the yaml that you can apply. So it should look like the second code snippet.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -166,6 +179,8 @@ func init() {
 	bootstrapCmd.Flags().BoolVar(&bootstrapCmdOptions.createDefaultRepository, "create-default-repository",
 		bootstrapCmdOptions.createDefaultRepository,
 		"Toggle creation of the default glasskube package repository")
+	bootstrapCmd.PersistentFlags().BoolVar(&bootstrapCmdOptions.dryrun, "dry-run", false,
+		"simulate the bootrapping of package without actually bootstrapping it")
```
christophenne, line 213:
```suggestion
		"simulate the bootstrapping of Glasskube without actually bootstrapping it")
```

pkg/bootstrap/bootstrap.go
```diff
@@ -132,6 +138,11 @@ func (c *BootstrapClient) Bootstrap(
 		manifests = append(manifests, defaultRepository())
 	}
 
+	if options.DryRun {
+		statusMessage("Dry-run mode enabled. The following resources would be applied:", true)
```
christophenne, line 142:
```suggestion
		statusMessage("Dry-run mode enabled. Glasskube Manifests will not be applied.", true)
```

christophenne (commented):
Please apply the two suggestions I left, I think then we can merge it.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -83,13 +84,24 @@ var bootstrapCmd = &cobra.Command{
 			fmt.Fprintf(os.Stderr, "\nAn error occurred during bootstrap:\n%v\n", err)
 			cliutils.ExitWithError()
 		}
-		if err := printBootsrap(
-			manifests,
-			bootstrapCmdOptions.Output,
-		); err != nil {
-			fmt.Fprintf(os.Stderr, "\nAn error occurred in printing : %v\n", err)
-			cliutils.ExitWithError()
+		if bootstrapCmdOptions.dryrun {
```
christophenne, line 87: The code if `dryrun` is true, and the code in the `else` is the same – or am I missing something? If not, please remove the duplicate branch, and on that occasion, please fix the typo `printBootsrap` -> `printBootstrap`. Thanks!

christophenne (commented):
Sorry for the late response, but I found one more thing I wanted to clarify. Thanks!

cmd/glasskube/cmd/bootstrap.go
christophenne: The `printBootstrap` function should be called in every case, sorry if that was unclear. It doesn't matter if `bootstrapCmdOptions.dryrun` is `true` or `false`, `printBootstrap` should be called anyway, just like it was before.

christophenne (changes requested):

kosmoz (changes requested):
This should use the unstructured clients dry-run feature. You can use it by setting the DryRun value in the Apply and Delete options to true

kosmoz (commented):
Please see my previous comment at https://github.com/glasskube/glasskube/pull/819#pullrequestreview-2144892558.

We want to utilize server-side dry run. `pkg/install/install.go` does something similar, so you can check that out if you're not sure what that means

hanshal101:
Ohk so referring the `pkg/install/install.go`. I tried implementing this code in the `bootstrap.go`
```go
createOptions := metav1.CreateOptions{}
	if options.DryRun {
		createOptions.DryRun = []string{metav1.DryRunAll}
		statusMessage("Dry-run mode enabled. Glasskube Manifests will not be applied.", true)
		return manifests, nil
	}
```
Instead of :
```go
if options.DryRun {
		statusMessage("Dry-run mode enabled. Glasskube Manifests will not be applied.", true)
		return manifests, nil
	}
```

Is this the correct approach?

kosmoz:
No. In the `applyManifests` manifest there is a call to `Apply`, where a `metav1.ApplyOptions` struct is already passed. You have to add the dry run option there.

https://github.com/glasskube/glasskube/blob/2d0c02b9183ef7b97f12e7afb47b2fd8bd516dfb/pkg/bootstrap/bootstrap.go#L223-L224

hanshal101:
Ohh, sorry I understood this. Rest all remains as it was previous.

kosmoz:
No worries.

Well, this part probably has to go:

https://github.com/glasskube/glasskube/blob/6d01d882ef61167156df39a0ac3bebf18d2b33ba/pkg/bootstrap/bootstrap.go#L142-L145

Then, the part where the status of workloads is checked must be skipped (because there no workloads are created with dry-run):

https://github.com/glasskube/glasskube/blob/6d01d882ef61167156df39a0ac3bebf18d2b33ba/pkg/bootstrap/bootstrap.go#L254-L260

Lastly, we have to do something about the job deletion, but I think we can figure that out as a step 2.

hanshal101:
```go
if !options.DryRun {
		for _, obj := range checkWorkloads {
			bar.Describe(fmt.Sprintf("Checking Status of %v (%v)", obj.GetName(), obj.GetKind()))
			if err := c.checkWorkloadReady(obj.GetNamespace(), obj.GetName(), obj.GetKind(), 5*time.Minute); err != nil {
				return err
			}
			_ = bar.Add(1)
		}
	}
```
Does this looks good. I called the `BootstrapOptions` in the `applymanifest` function.

kosmoz:
Yes that should work

hanshal101:
well I tried this in my local env, output looks something like this.
```powershell
PS H:\glasskube> .\main.exe bootstrap --dry-run
Glasskube will be updated to version v0.11.0 in cluster kind-lol.
Continue? (Y/n) y

## Installing GLASSKUBE ##
🧊 The next generation Package Manager for Kubernetes 📦
* Fetching Glasskube manifest from https://github.com/glasskube/glasskube/releases/download/v0.11.0/manifest-aio.yaml
* Validating existing installation
* Applying Glasskube manifests
Applying glasskube (PackageRepository)  94% |███████████████████████████████████████████████████████████████████████████████████████████      | (49/52, 5 it/s) [10s:0s]* Telemetry is enabled for this cluster – Run "glasskube telemetry status" for more info.
* Glasskube successfully installed! (took 21s)
```

kosmoz:
That's great, it means nothing was applied in the cluster, which is exactly what we want for a dry-run.

I think we should also add a log line confirming that dry run is enabled at the start: 

> Dry-run is enabled. Nothing will be changed in your cluster!

hanshal101:
Well I have now added the `statusMessage("Dry-run mode enabled. Glasskube Manifests will not be applied.", true)` here.

pkg/bootstrap/bootstrap.go
```diff
@@ -221,7 +233,14 @@ func (c *BootstrapClient) applyManifests(ctx context.Context, objs []unstructure
...
 		if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
-				Apply(ctx, obj.GetName(), &obj, metav1.ApplyOptions{Force: true, FieldManager: "glasskube"})
+				Apply(
+					ctx, obj.GetName(),
+					&obj,
+					metav1.ApplyOptions{
+						Force:        true,
+						FieldManager: "glasskube",
+						DryRun:       []string{metav1.DryRunAll},
```
kosmoz, line 242: The `DryRun` property should only be set if dry-run is enabled via options

kosmoz (changes requested):

pkg/bootstrap/bootstrap.go
hanshal101:
```go
var dryrun []string
		if options.DryRun {
			dryrun = []string{metav1.DryRunAll}
		}

		if err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
			_, err = c.client.Resource(mapping.Resource).Namespace(obj.GetNamespace()).
				Apply(
					ctx, obj.GetName(),
					&obj,
					metav1.ApplyOptions{
						Force:        true,
						FieldManager: "glasskube",
						DryRun:       dryrun,
					})
			return err
		}); err != nil {
			return err
		}
		```
		
		Something like this would do?

kosmoz: yes something like that

kosmoz (approved):

christophenne (approved):

PR #822, opened 2024-06-18
feat(cli): bootstrap shows different prompt for bootstrapped clusters

### LINKED ISSUE #746: `glasskube bootstrap` should show different prompt when updating
Currently the `glasskube bootstrap` command shows a prompt like "Glasskube will be installed in context xyz. Continue?" (see #719). While this is correct for non-bootstrapped clusters, it makes no sense when the user wants to update the current glasskube installation and might be confusing. 

When glasskube is bootstrapped in the current cluster, the prompt should be: "Glasskube will be updated to version [version] in cluster [cluster]. Continue?" 

As in #719, if the `--yes` flag is set, the prompt should not appear. 

There is already a check in place, whether the installation would be downgraded. This check should be done first. If the installation would not be downgraded, then the prompt for installing/updating should appear, because only at this point we know if glasskube is already installed and in which version. 

**Additional context**

utsab818:
Closes #746 

## 📑 Description
`glasskube bootstrap` command for bootstrapped-clusters now give updated prompt to avoid confusion.
Most cases when upgrading the version is handled with proper prompt.

## ✅ Checks


christophenne (commented):
Thanks for your PR!

Looks good, I noticed two things when testing, see comments. Please let me know if you want to incorporate these yourself or if we should open up another issue.

utsab818:
@christophenne I will definitely try to incorporate those lines of codes.

utsab818:
@christophenne Does this look good or is there any other cases or any other proper methods to do so?
image

christophenne:
@utsab818 the check for installing the same version doesn't work, because `semver.IsUpgradable(installedVersion, desiredVersion)` returns false when the two versions are the same. So this check should be outside of the `upgradeNeeded` if, I think. The rest works fine. 

Testing hint: 
* `minikube start -p bootstrap-test`
* change `var Version` in `internal/config.go` to `"0.8.0"` (for example)
* `go run cmd/glasskube/main.go bootstrap`

Then you can change the version as you wish and test all the cases. When done, delete the minikube cluster with `minikube delete -p bootstrap-test`.

utsab818:
@christophenne Thanks for the hint. I will update it.

christophenne:
> @christophenne Thanks for the hint. I will update it.

Thanks! However, now bootstrap doesn't work in an empty cluster, I think you should additionally check for `installedVersion != ""` in that `if`: 

```
panic: runtime error: slice bounds out of range [1:0]

goroutine 1 [running]:
github.com/glasskube/glasskube/cmd/glasskube/cmd.init.func1(0x3221140, {0x32a2300?, 0x4?, 0x1e63fed?})
        /home/christoph/dev/glasskube/cmd/glasskube/cmd/bootstrap.go:99 +0x92e
github.com/spf13/cobra.(*Command).execute(0x3221140, {0x32a2300, 0x0, 0x0})
```

utsab818:
@christophenne Sorry for causing you trouble with these small things. 
I have added installedVersion != "" in elseif.
Is there any other criteria that needs to be looked for?
I tried doing some tests and seems to be working fine.

cmd/glasskube/cmd/bootstrap.go
```diff
@@ -82,6 +75,43 @@ var bootstrapCmd = &cobra.Command{
...
+					}
+				} else {
+					confirmMessage := fmt.Sprintf("Glasskube will be updated to version %s "+
+						"in cluster %s.\nContinue? ", desiredVersion, currentContext)
+					if !cliutils.YesNoPrompt(confirmMessage, true) {
+						fmt.Println("Operation stopped")
+						cliutils.ExitWithError()
+					}
+				}
+			} else if installedVersion[1:] == desiredVersion {
```
christophenne, line 99: `installedVersion` could be empty here, and then the `[1:]` access panics. So this condition should be `installedVersion != "" && ...`

christophenne:
> @christophenne Sorry for causing you trouble with these small things. I have added installedVersion != "" in elseif. Is there any other criteria that needs to be looked for? I tried doing some tests and seems to be working fine.

No problem, did you maybe not commit/push it?

The test case I tried is simply a bootstrap in a new empty cluster. This will lead to `installedVersion` being empty.

utsab818:
@christophenne I will push it now. I was just waiting for your confirmation.

christophenne (approved):

PR #824, opened 2024-06-18
feat: add verifying breaking changes during bootstrap

kosmoz:
Ref #803

## 📑 Description
This PR adds a mechanism to prevent using bootstrap for upgrades that include breaking changes. 

## ✅ Checks

## ℹ Additional Information
Related docs PR: https://github.com/glasskube/glasskube/pull/825


pkg/bootstrap/bootstrap.go
```diff
@@ -147,6 +153,35 @@ func (c *BootstrapClient) Bootstrap(
 	return manifests, nil
 }
 
+func (c *BootstrapClient) verifyLegalUpdate(ctx context.Context, targetVersion string) error {
```
christophenne, line 156: Since there are already a few bootstrap version checks in place in the command, do you think it would make sense to move this over there as well? I'm sure all of this logic could be combined into some dedicated module (doesn't have to be now), which could be later on used on the UI if we ever want to do upgrades there as well.

christophenne (approved):
LGTM

PR #827, opened 2024-06-19
fix(cli): `repo add --default` removes annotation for current default repo

### LINKED ISSUE #806: `repo add` command with --default flag should return an error if there's a repository with that annotation already
**Describe the bug**
[Documentation](https://glasskube.dev/docs/design/repositories/#repository-management) states that if you use --default with the `repo add` command it should return an error if there's a repository with that annotation already.

**To Reproduce**
1. Try to add a new repository with `glasskube repo add test https://example.com --default`
2. List repositories with `glasskube repo list`
3. Check that there are two default repositories

**Expected behavior**
Either we should get an error, or instead the current default repo will get the default annotation removed and the new repository being added will be the new default repo

**Screenshots**

image

 - Type: Kubernetes
 - Version: 1.30.0
 - Provider: minikube

 - Package Operator version: v0.0.1
 - CLI version: dev

doncicuto:
Closes #806 

## 📑 Description

This PR changes `repo add --default` so we can only have one default repository

## ✅ Checks
- [X] I have updated the documentation as required

## ℹ Additional Information

If we now add a new repo using the --default flag we can have two or more repositories with the default annotation set. With this PR the command updates the current default repo to remove the default annotation so we only have one default repository. If we've an error when adding the new repo we roll back and set the default annotation for the previous repo.

### DIFF
--- a/cmd/glasskube/cmd/repo_add.go
+++ b/cmd/glasskube/cmd/repo_add.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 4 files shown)


kosmoz (changes requested):
Looks good, but this fails if there is currently no default repository. I think the best way to solve this would be a sentinel error: https://www.digitalocean.com/community/tutorials/how-to-add-extra-information-to-errors-in-go#handling-specific-errors-using-sentinel-errors (but use `errors.Is()` for comparison instead of `==`)

doncicuto:
Hi @kosmoz, thanks for the hint, I've added the sentinel error for the no default repo found issue in which case the flow continue to create/update the repository

kosmoz (approved):

PR #837, opened 2024-06-21
fix(ui): open package description links in new tab

### LINKED ISSUE #832: Links in package description should open in new tab
The `longDescription` of a package manifest can contain links in markdown format. A markdown link is correctly rendered as an html anchor element, but it opens in the same tab the user is currently in.

Links in the `longDescription` should be rendered to have the `target="_blank"` attribute too, to make it open in a new tab.

**Additional context**
This description is converted from markdown to HTML on the `package.html` template, using the `| Markdown` pipe. The corresponding function `Markdown` is registered in the `templates.go` file. There, the rendering happens – probably here something has to be done.

doncicuto:
Closes #832 

## 📑 Description

This PR makes links in package description open in a new tab: `target="_blank" rel="noopener noreferrer"`

## ✅ Checks

## ℹ Additional Information

This PR uses https://pkg.go.dev/golang.org/x/net/html to parse the html and manipulate the nodes to add the target and rel attributes to any link found in the converted markdown. Here's a vide showing the new behavior:

https://github.com/glasskube/glasskube/assets/30386061/fbb2c376-ceaf-45f4-a9f7-b96aaac58c2e

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/spf13/cobra v1.8.1
 	github.com/yuin/goldmark v1.7.2
 	go.uber.org/multierr v1.11.0
+	golang.org/x/net v0.25.0
 	golang.org/x/term v0.21.0
 	k8s.io/api v0.30.2
 	k8s.io/apiextensions-apiserver v0.30.2
@@ -83,7 +84,6 @@ require (
 	github.com/wk8/go-ordered-map/v2 v2.1.8 // indirect
 	go.uber.org/zap v1.26.0 // indirect
 	golang.org/x/exp v0.0.0-20240112132812-db7319d0e0e3 // indirect
-	golang.org/x/net v0.25.0 // indirect
 	golang.org/x/oauth2 v0.16.0 // indirect
 	golang.org/x/sys v0.21.0 // indirect
 	golang.org/x/text v0.15.0 // indirect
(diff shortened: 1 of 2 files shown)


kosmoz (commented):
I think it should be possible to achieve this during markdown rendering. 

I found these related resources:
- https://github.com/yuin/goldmark/issues/160
- https://github.com/go-gitea/gitea/blob/814ca9ffea229f55f1f69b34d9a7f84054115669/modules/markup/markdown/goldmark.go#L110
- https://github.com/yuin/goldmark/issues/25

You can create an ASTTransformer like so:

```go
type ASTTransformer struct{}

func (*ASTTransformer) Transform(node *ast.Document, reader text.Reader, pc parser.Context) {
	_ = ast.Walk(node, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
		// Do something with the node
		return ast.WalkContinue, nil
	})
}
```

And then add it to the converter like so:

```go
converter := goldmark.New(
	goldmark.WithParserOptions(
		parser.WithASTTransformers(
			util.Prioritized(&ASTTransformer{}, 1000),
		),
	),
)
converter.Convert(/* ... */)
```

doncicuto:
Thanks @kosmoz I've tried to apply your proposed approach working directly with the markdown

kosmoz (approved):
Awesome, works like a charm! :tada:

Issue #838, opened 2024-06-21
Server out of sync with Kubernetes API server


christophenne:
**Describe the bug**
At random times, the watches of the UI server to the API server are "stuck", i.e. no events are received anymore, but there is no error or anything indicating so. As a consequence the UI does not react anymore to any cluster updates. 

**To Reproduce**
No actual reproduction scenario found yet. Sometimes happens after 1 minute, sometimes after 4 hours. 

**Expected behavior**
When watches are lost, they have to be reconnected, and if that fails or is not possible, the user needs to be notified somehow about the error. 

**Screenshots**

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: 1.28, 1.29, 1.30
 - Provider: minikube, OVH

**Glasskube Info (please complete the following information):**
 - Package Operator version: all recent versions
 - CLI version: all recent versions

**Additional context**
Related: https://github.com/glasskube/glasskube/issues/768, https://github.com/glasskube/glasskube/issues/756

christophenne:
Potential lead: https://github.com/glasskube/glasskube/pull/1027#issuecomment-2244662997 (could be that some cache is not updating anymore at some point?)

PR #843, opened 2024-06-24
fix(client): propagate list options and apply timeout

christophenne:
Ref #838 

## 📑 Description
This is an attempt at making the out of sync problem less likely to happen, by passing a fixed timeout in the API call (like also all standard listers for namespace, secrets, etc do). The "base" timeout is set by the lister/watcher of the go client, but we divide it by 10 to "force" rewatches in smaller iterations. This should lead to re-watches every 30 - 60 seconds, so even if the out of sync issue appears, it would hopefully repair itself after at latest 1 minute. 

Additionally, by passing the given list options to our client, and reusing it rather than overwriting it, we will set the `allowWatchBookmarks` flag to true. Therefore we will also process `BOOKMARK` events, which might have an influence on the issue: https://stackoverflow.com/questions/66080942/what-k8s-bookmark-solves

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -1004,7 +1004,7 @@ func (s *server) initPackageStoreAndController(ctx context.Context) (cache.Store
 				return &pkgList, err
 			},
 			WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-				return pkgClient.ClusterPackages().Watch(ctx)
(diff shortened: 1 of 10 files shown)


kosmoz (commented):
I have one suggestion.

Apart from that, this looks like a good thing to try.

kosmoz (approved):

PR #851, opened 2024-06-24
feat(cli): add installing namespace-scoped packages

kosmoz:
Ref #803

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -11,6 +11,7 @@ import (
 	"github.com/glasskube/glasskube/internal/clicontext"
 	"github.com/glasskube/glasskube/internal/cliutils"
 	"github.com/glasskube/glasskube/internal/config"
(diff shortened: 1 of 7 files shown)


cmd/glasskube/cmd/install.go
```diff
@@ -118,15 +119,48 @@ var installCmd = &cobra.Command{
...
+			if len(args) != 2 {
+				fmt.Fprintf(os.Stderr, "%v has scope Namespaced. Please enter a name (default %v):\n", packageName, packageName)
+				name = cliutils.GetInputStr("name")
+				if name == "" {
+					name = packageName
+				}
+			} else {
+				name = args[1]
+			}
+			ns := installCmdOptions.GetActualNamespace(ctx)
```
christophenne, line 153: Installing with the `--namespace` argument does not use the given namespace, because `.GetActualNamespace` is accessing `configureCmdOptions`.

PR #875, opened 2024-06-26
feat(ui): introduce cache verifier

christophenne:
Ref #838 

## ✅ Checks


kosmoz (changes requested):
I have some thoughts on the general architecture, but the feature works nicely :slightly_smiling_face:

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,250 @@
...
+				fmt.Fprintf(os.Stderr, "CACHEVERIFIER: failed get cached items: %v\n", err)
+			} else {
+			outerLoop:
+				for _, nonCachedItem := range nonCachedItems {
+					for _, cachedItem := range cachedItems {
+						if nonCachedItem.GetNamespace() == cachedItem.GetNamespace() &&
+							nonCachedItem.GetName() == cachedItem.GetName() {
+							if nonCachedItem.GetName() == "glasskube" {
+								nonCachedItem.SetResourceVersion("asdf")
+							}
```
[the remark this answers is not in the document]
christophenne: yes thanks, that was for testing and i forgot to remove it :smile:

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,250 @@
...
+				"(cached resource version is '%s', but expected '%s')", diffMapKey, cachedResourceVersion, nonCachedResourceVersion)
+		}
+		// else: cached item has been updated since last time, but is not yet update, which is okay
+		// we consider this as not being stuck (yet), but mark for check in next round
+	}
+	// mark for check in the next verification round
+	verifier.diffMap[diffMapKey] = diffMapEntry{
+		cachedResourceVersion:    cachedResourceVersion,
+		nonCachedResourceVersion: nonCachedResourceVersion,
+	}
```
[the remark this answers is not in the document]
christophenne: Good catch, thanks

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,250 @@
...
+	clear(verifier.diffMap)
+	verifier.errCh <- err
+}
+
+func (verifier *cacheVerifier) newStopCh() chan struct{} {
+	verifier.stopCh = make(chan struct{})
+	return verifier.stopCh
+}
+
+func (verifier *cacheVerifier) handleDiff(cachedObj metav1.Object, nonCachedObj metav1.Object) error {
```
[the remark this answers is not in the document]
christophenne: True, any error would have been considered as a failed verification, while it could be an "actual" error. I changed it to a custom error type now, and check for that. 
The idea of using an error was to transport detailed information to the outside (e.g. logs and telemetry), and from my perspective that should be reasonable: We do something similar when we have http errors for example, we also check for the status code there and make our subsequent logic dependent on it. Same for our configuration validations. But maybe I am missing something?

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,225 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+type cacheVerifier struct {
+	cachedClient    client.PackageV1Alpha1Client
+	nonCachedClient client.PackageV1Alpha1Client
+	lister          verifyLister
+	mutex           sync.Mutex
+	diffMap         map[string]diffMapEntry
```
kosmoz, line 24: Maybe use a sync.Map instead?

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,225 @@
...
+
+	var diffErr *cacheVerificationError
+
+endlessLoop:
+	for range tickC {
+		if nonCachedItems, err := verifier.listNonCachedItems(ctx); err != nil {
+			fmt.Fprintf(os.Stderr, "CACHEVERIFIER: failed to get actual items: %v\n", err)
+		} else if cachedItems, err := verifier.listCachedItems(ctx); err != nil {
+			fmt.Fprintf(os.Stderr, "CACHEVERIFIER: failed get cached items: %v\n", err)
+		} else {
```
kosmoz, line 119: by using `verifier.lister(ctx, cachedClient)` and `verifier.lister(ctx, nonCachedClient)` here, we could avoid storing the clients in the struct state at all. What do you think about that?

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,210 @@
...
+			return errors.New("unexpected cache type in diffMap")
+		}
+		// else: cached item has been updated since last time, but is not yet update, which is okay
+		// we consider this as not being stuck (yet), but mark for check in next round
+	}
+	// mark for check in the next verification round
+	verifier.diffMap.Store(diffMapKey, diffMapEntry{
+		cachedResourceVersion:    cachedResourceVersion,
+		nonCachedResourceVersion: nonCachedResourceVersion,
+	})
```
kosmoz, line 208: please use LoadOrStore because otherwise it is not atomic

internal/web/cacheverifier.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	verifier.closeAndReset(diffErr)
+}
+
+func (verifier *cacheVerifier) closeAndReset(err error) {
+	close(verifier.stopCh)
+	verifier.diffMap.Range(func(key, value any) bool {
+		verifier.diffMap.Delete(key)
+		return true
+	})
```
kosmoz, line 166: alternatively you could just reset to a new empty `sync.Map{}` let the GC do the rest

Issue #850, opened 2024-06-24
Add asterisks to input label for required configuration values


pmig:
**Is your feature request related to a problem? Please describe.**
It is not easily possible to determine if an configuration value is mandatory.

**Describe the solution you'd like**
If a configuration value as marked as mandatory an asteriks should be added to the label

**Describe alternatives you've considered**
None, open for suggestions

**Additional context**
 - Bootstrap 5 related documentation: https://getbootstrap.com/docs/5.3/forms/validation/
 - Example solution of an asterisk: https://stackoverflow.com/questions/23141854/adding-asterisk-to-required-fields-in-bootstrap-3
 - related to https://github.com/glasskube/glasskube/issues/849

rharshit82:
Can you please assign me this ?

Issue #849, opened 2024-06-24
Native html validation seems to be broken


pmig:
**Describe the bug**
Html form validation is broken

**To Reproduce**
1. open the ui `glasskube open`
2. configure a package with a mandatory configuration values (e.g. `k8sgpt-operator`)
3. save a package without entering a mandatory value

**Expected behavior**
Html validation should work

**Screenshots**
![image](https://github.com/glasskube/glasskube/assets/3041752/0e592624-c600-4ad9-a644-7bd3493bc992)

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: 1.30.0
 - Provider: minikube

**Glasskube Info (please complete the following information):**
```
at 14:29:49 ➜ glasskube version
glasskube: v0.10.0
package-operator: v0.10.0
```

**Additional context**
- This is a regression.
- The required attribute should be added if a configuration value is marked as required

christophenne:
Some technical context:

* you can find all the input templates for the configuration page in `internal/web/templates/components/pkg-config.input.html`
* required validation works for html option elements (dropdown), see `pkg-config-input-options`
* it's the same for number and text inputs, but for some reason just missing there. it should be added to these two input types too.

rharshit82:
Have added this in the same PR that added asterisks. I found this bug while doing review changes.

PR #904, opened 2024-06-27
feat(ui): added asterisk for mandatory form inputs

rharshit82:
Closes https://github.com/glasskube/glasskube/issues/850
Closes https://github.com/glasskube/glasskube/issues/849

## 📑 Description
Added asterisks to mandatory inputs if constraint is defined for that particular input. 

## ✅ Checks

### DIFF
--- a/internal/web/templates/components/pkg-config-input.html
+++ b/internal/web/templates/components/pkg-config-input.html
@@ -225,12 +225,17 @@
 {{ end }}
 
 
-<!-- this is the entry point for rendering the the input group for one value definition -->
+<!-- this is the entry point for rendering the input group for one value definition -->
 {{ define "pkg-config-input" }}
   <div id="{{ .ContainerId }}" class="mb-2">
     {{ if eq .ValueDefinition.Type "text" }}
       <div>
-        <label for="{{ .FormId }}" class="form-label">{{ .FormLabel }} (<code>{{ .ValueName }}</code>)</label>
+        <label for="{{ .FormId }}" class="form-label"
+          >{{ .FormLabel }}{{ if .ValueDefinition.Constraints.Required }}
+            &nbsp;<span class="text-danger">*</span>
+          {{ end }}
+          (<code>{{ .ValueName }}</code>)</label
+        >
         <div class="input-group input-group-sm">
           {{ template "pkg-config-input-reference-dropdown" . }}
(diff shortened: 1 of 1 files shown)


internal/web/templates/components/pkg-config-input.html
```diff
@@ -225,12 +225,17 @@
...
 
 
-<!-- this is the entry point for rendering the the input group for one value definition -->
+<!-- this is the entry point for rendering the input group for one value definition -->
 {{ define "pkg-config-input" }}
   <div id="{{ .ContainerId }}" class="mb-2">
     {{ if eq .ValueDefinition.Type "text" }}
       <div>
-        <label for="{{ .FormId }}" class="form-label">{{ .FormLabel }} (<code>{{ .ValueName }}</code>)</label>
+        <label for="{{ .FormId }}" class="form-label"
```
christophenne, line 233: Since it's the same code snippet for all types of inputs, I would recommend doing a little template for that. 

You can add something like 
```
{{ define "pkg-config-input-required" }}
 ...
{{ end }}
```

and then render this template with `{{ template "pkg-config-input-required" . }}` 4 times.

internal/web/templates/components/pkg-config-input.html
```diff
@@ -225,12 +225,17 @@
...
+          >{{ .FormLabel }}{{ if .ValueDefinition.Constraints.Required }}
+            &nbsp;<span class="text-danger">*</span>
```
christophenne, line 235: IMO you can get rid of the `&nbsp;`

christophenne (changes requested):
@rharshit82 thanks for taking this, I have on small comment and then we can merge it :)

internal/web/templates/components/pkg-config-input.html
rharshit82: done

internal/web/templates/components/pkg-config-input.html
rharshit82: done

rharshit82 (commented):
This PR fixes https://github.com/glasskube/glasskube/issues/849 as well. Found the issue while doing review changes. I'll rebase this PR once it gets approved.

christophenne (approved):
Thanks for fixing the bug as well, good job!

Issue #905, opened 2024-06-28
Auto-Update is not shown on discussion page


christophenne:
**Describe the bug**
The package detail page and package discussion page share the same header template, which includes logo, name, and the badges showing installation information (version, repo, auto-update). However, on the discussion page, the auto-update value is empty, while on the package detail page it say "Disabled" or "Enabled". 

**To Reproduce**
1. Go to the package detail page of an installed package
2. Click on the discussion badge to get to the discussion page

**Expected behavior**
Auto update value should be the same as on package detail page. 

**Screenshots**

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: 1.30
 - Provider: minikube

**Glasskube Info (please complete the following information):**
 - Package Operator version: v0.10.1
 - CLI version: dev (not released yet)

**Additional context**

rharshit82:
Since this was a minor change and very less time passed since the issue was raised, didn't ask earlier. Can you please assign me this?

christophenne:
Sure, thanks, I'll take a look soon.

PR #907, opened 2024-06-28
fix(ui): autoupdate label for discussion page

rharshit82:
Closes https://github.com/glasskube/glasskube/issues/905

## 📑 Description
auto update value was not passed to the discussion page whereas it was passed to package detail page, hence the bug.

## ✅ Checks

### DIFF
--- a/internal/web/discussions.go
+++ b/internal/web/discussions.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/glasskube/glasskube/internal/clientutils"
 	"github.com/glasskube/glasskube/internal/web/util"
 
 	"github.com/glasskube/glasskube/internal/giscus"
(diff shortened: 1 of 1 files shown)


internal/web/discussions.go
```diff
@@ -103,6 +123,7 @@ func (s *server) handlePackageDiscussionPage(w http.ResponseWriter, r *http.Requ
 		"ShowDiscussionLink": true,
 		"PackageHref":        pkgHref,
 		"DiscussionHref":     fmt.Sprintf("%s/discussion", pkgHref),
+		"AutoUpdate":         d.AutoUpdate,
```
christophenne, line 126: The auto update value only depends on the `pkg`, which is already contained in the `packageDetailPageContext`. So just as in the package detail pages, it would be easier to use `clientutils.AutoUpdateString(pkg)` here, which also takes care if `pkg` is nil.

christophenne (commented):
Thanks! I like your idea with the new context (and that's how it could be done in case of additional properties only for the discussion page), but I don't think it's necessary in this case, see the comment.

internal/web/discussions.go
rharshit82: ahh nice, missed this earlier. done.

PR #911, opened 2024-06-28
fix(ui): show an error if the client cache is out of sync

### LINKED ISSUE #768: Wait for server caches to be populated
Turning off some logging shows that we open the browser and show the "ready message" ("UI is available ...") a bit too early: 

```
I0606 15:14:07.129073  279189 cert_rotation.go:137] Starting client certificate rotation controller
I0606 15:14:07.135934  279189 reflector.go:296] Starting reflector *v1alpha1.Package (0s) from pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232
I0606 15:14:07.135937  279189 reflector.go:296] Starting reflector *v1alpha1.PackageInfo (0s) from pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232
...
I0606 15:14:07.137386  279189 reflector.go:359] Caches populated for *v1.Namespace from pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232
I0606 15:14:07.137814  279189 reflector.go:359] Caches populated for *v1alpha1.PackageInfo from pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232
I0606 15:14:07.137825  279189 reflector.go:359] Caches populated for *v1alpha1.Package from pkg/mod/k8s.io/client-go@v0.30.1/tools/cache/reflector.go:232
...
```

When the connection to the cluster is slow, this could lead to an "empty" (i.e. nothing installed, even though packages are installed) package overview page on the initial load – happened to me once in my OVH cluster. It was corrected after page reload, so I suspect this timing issue was the cause of that. 

It should be investigated whether we can actually hook into the store/controller of `client-go` (or maybe we can somehow use our `initPackageStoreAndController` functions?), or not at all – if not, I guess the issue can be neglected.
If it's possible, we could 
* send an update event to the client once caches are populated, then it will rerender anyway,
* show a warning on the UI: "server is initializing" or something like that,
* or not open the browser until caches are populated (but of course one can still always open the page in the browser by themselves), 
* or both. 

**Additional context**

christophenne:
Ref #838
Closes #768 
Undoes changes of #843

## 📑 Description
As the reconnection logic of #875 does not work yet, we want to provide users with a log message in the server, and shut it down so that it's necessary to restart manually again. This is of course far from ideal, but the best/cheapest way to tell the user something is wrong (the UI will show the "disconnected" banner when the server is down). 

## ✅ Checks

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -247,3 +247,9 @@ func SetUserProperty(key string, value string) {
 		},
 	})
 }
+
+func ReportCacheVerificationError(err error) {
+	ev := instance.getBaseEvent("error_cache_verification", false)
+	ev.Properties["err"] = err.Error()
+	_ = instance.posthog.Enqueue(ev)
+}
--- a/internal/web/cacheverifier.go
+++ b/internal/web/cacheverifier.go
@@ -1,0 +1,214 @@
+package web
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/rand"
+	"os"
+	"sync"
+	"time"
+
+	"github.com/glasskube/glasskube/api/v1alpha1"
+	"github.com/glasskube/glasskube/pkg/client"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+)
+
+type cacheVerifier struct {
+	lister     verifyLister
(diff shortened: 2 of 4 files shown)


kosmoz (approved):

kosmoz:
Note that I wasn't able to reproduce the bug this time, but the code looks good :+1:

PR #921, opened 2024-07-01
chore(ui): telemetry should exclude certain paths

christophenne:
## ✅ Checks

### DIFF
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/telemetry/client.go
```diff
@@ -214,18 +215,25 @@ func (t *ClientTelemetry) close() {
...
 }
 
-func HttpMiddleware() func(http.Handler) http.Handler {
+type PathRedactor = func(url string) string
+
+type HttpMiddlewareOptions struct {
+	PathRedactors []PathRedactor
+}
+
+func HttpMiddleware(options HttpMiddlewareOptions) func(http.Handler) http.Handler {
```
kosmoz, line 224: A pattern I see a lot for this kind of thing (options parameter) is the use of configuration functions (let me know if you think this would make sense here). It would look something like this:
```suggestion
type httpMiddlewareOptions struct {
	pathRedactors []PathRedactor
}

type HttpMiddlewareConfigurator = func(opt *httpMiddlewareOptions)

func WithPathRedactor(pr PathRedactor) HttpMiddlewareConfigurator {
  return func(opt *httpMiddlewareOptions) {
    opt.pathRedactors = append(opt.pathRedactors, pr)
  }
}

func HttpMiddleware(conf HttpMiddlewareConfigurator...) func(http.Handler) http.Handler {
  var options httpMiddlewareOptions
  for _, c := range conf {
    c(&options)
  }
```

Then in `server.go` it would look something like this:

```go
router.Use(telemetry.HttpMiddleware(telemetry.WithPathRedactor(packagesPathRedactor)))
```

kosmoz: I guess it's a bit like a "builder" pattern but in a more functional style

kosmoz: Examples:
- https://pkg.go.dev/github.com/schollz/progressbar/v3#Option
- https://pkg.go.dev/github.com/kaleocheng/goldmark#Option

christophenne: Interesting, looks good to me :+1:

kosmoz (approved):

PR #929, opened 2024-07-02
fix(cli): `--outdated` and `--installed` for listing packages

christophenne:
## ✅ Checks

### DIFF
--- a/pkg/list/lister.go
+++ b/pkg/list/lister.go
@@ -64,7 +64,7 @@ func (l *lister) GetClusterPackagesWithStatus(
 	index, err := l.fetchRepoAndInstalled(ctx, options, includeClusterPackages)
 	result := make([]*PackageWithStatus, 0, len(index))
 	for _, item := range index {
-		if itemShouldBeIncluded(&item, options) {
+		if clusterPackageShouldBeIncluded(&item, options) {
 			pkgWithStatus := PackageWithStatus{
 				MetaIndexItem:  *item.IndexItem,
 				ClusterPackage: item.ClusterPackage,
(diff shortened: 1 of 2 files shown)


pkg/list/lister.go
```diff
@@ -86,31 +86,44 @@ func (l *lister) GetPackagesWithStatus(
...
 			}
-			ls = append(ls, &pkgWithStatus)
 		}
-		result = append(result, &PackagesWithStatus{
-			MetaIndexItem: *item.IndexItem,
-			Packages:      ls,
-		})
-		// }
+		hasIncludedItems := len(ls) > 0
+		if hasIncludedItems || (!hasIncludedItems && !options.OnlyInstalled && !options.OnlyOutdated) {
```
kosmoz, line 104: I think this can be simplified
```suggestion
		if hasIncludedItems || (!options.OnlyInstalled && !options.OnlyOutdated) {
```

pkg/list/lister.go
```diff
@@ -86,31 +86,44 @@ func (l *lister) GetPackagesWithStatus(
...
 
-func itemShouldBeIncluded(item *result, options ListOptions) bool {
-	return !((options.OnlyInstalled && !item.Installed()) || (options.OnlyOutdated && !item.Outdated()))
+func clusterPackageShouldBeIncluded(item *result, options ListOptions) bool {
+	return !((options.OnlyInstalled && !item.ClusterPackageInstalled()) ||
+		(options.OnlyOutdated && !item.ClusterPackageOutdated()))
+}
+
+func packageShouldBeIncluded(item *result, pkg ctrlpkg.Package, options ListOptions) bool {
+	if !pkg.IsNil() {
```
kosmoz, line 120: I think this function should only be called with a `*v1alpha1.Package`, since there is also a function called `clusterPackageShouldBeIncluded`, so it would make sense to make that explicit via the parameter type
```suggestion
func packageShouldBeIncluded(item *result, pkg *v1alpha1Package, options ListOptions) bool {
	if pkg != nil {
```

kosmoz (changes requested):

kosmoz (approved):

PR #937, opened 2024-07-02
feat(ui): support partial rerender for package details

### LINKED ISSUE #729: Package Config Form should not be reset when only package status updates
We rerender the whole package detail page at any package update. If a user is working in the package configuration and changes a few values without having saved yet, and package update is coming in, the page is being rerendered and the changes are lost. 

The config form should not be reset when only the status changes. It is probably fine to still reset it in these cases:
* package has been installed or uninstalled
* package configuration has changed elsewhere
* package version/repository has changed

Old and new package are both known at the time of the update, so the config should be comparable. Since the package detail page is quite bloated already anyway, it might make sense to somehow split it into separate components/endpoints, which would have the benefit that separate components can be reloaded instead of the whole page. 

**Additional context**

christophenne:
Closes #729

## 📑 Description
In this PR, partial rerendering of the package detail header is added. For that, the server endpoint takes an additional query param `"component"` which can be empty (for full page) or `"header"`. On the client, there is an additional `sse:` trigger now, that refreshes only the header when the corresponding trigger ID is received via SSE. 
To identify whether only the header or the full package detail page have to be reloaded, the old and new package are compared with `reflect.DeepEqual`.

There is also some refactoring:
* handling of all of the SSE trigger IDs has been streamlined into one util file.
* handling of broadcast updates by introducing a `broadcaster` that handles the low-level SSE hub + checks the diff of two objects and sends the corresponding SSE triggers downstream. 
* `sseHub` was changed to use a `sync.Map`

## ✅ Checks


internal/web/sse/ssehub.go
```diff
@@ -0,0 +1,93 @@
...
+	// unregister requests from clients.
+	unregister chan *sseClient
+
+	// Registered clients.
+	clients sync.Map // map[*sseClient]struct{}
+}
+
+type sse struct {
+	event string
+	data  string
```
kosmoz, line 28: I noticed that this is never used. can it be removed?

kosmoz (commented):
Please have a look at this comment. Apart from that, it works great and it looks like this includes a good bit of cleanup too, so that's nice :broom:

internal/web/sse/ssehub.go
christophenne: It's used in the `sseHub` struct for the broadcasting channel, and `sse` objects are also created in the `broadcaster`

kosmoz: Sorry, I was referring only to the `data` property.

christophenne: Oh, good point. Originally we sent the rendered html in the `data` property, but currently we don't use it because we only use the `event` to define the trigger. 
I'm unsure, but I would still leave it in, because otherwise that would be an SSE Hub that couldn't send actual data.

kosmoz: OK, let's keep it then. I just wanted to bring it up

kosmoz (approved):

PR #938, opened 2024-07-02
feat(cli): support for changing configuration on package update

### LINKED ISSUE #864: CLI: Add support for changing configuration during update
If the value definitions of a package are different between versions, it is sometimes not possible to update safely. Particularly, if a new required value is added, it is not possible to update at all.

When running `glasskube update` to update a package with configuration values, the user should be asked if they want to change the configuration. If they answer yes, the configuration wizard should be started.

Harshitk-cp:
Closes #864

## 📑 Description
added configuration upgrade support on `glasskube update`

## ✅ Checks

## ℹ Additional Information


pkg/manifest/manifest.go
```diff
@@ -38,3 +39,33 @@ func GetInstalledManifestForPackage(ctx context.Context, pkg ctrlpkg.Package) (*
...
+				&manifest,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("error fetching manifest: %w", err)
+			}
+			return &manifest, nil
+		}
+	}
+
+	return nil, fmt.Errorf("version %s not found for package %s", version, pkg.GetSpec().PackageInfo.Name)
```
kosmoz, line 70: Please correct me if I'm wrong, but I think this is unnecessarily complex. 
```suggestion
	var manifest v1alpha1.PackageManifest
	err := repoClient.ForPackage(pkg).FetchPackageManifest(
		pkg.GetSpec().PackageInfo.Name,
		version,
		&manifest,
	)
	if err != nil {
		return nil, fmt.Errorf("error fetching manifest: %w", err)
	}
	return &manifest, nil
```

kosmoz (changes requested):

kosmoz (approved):

PR #939, opened 2024-07-02
feat(cli): added status message to describe command

### LINKED ISSUE #927: Display status message in `glasskube describe` command
Like #899, but for the describe command of a package.

Message should be shown below the status, but non-installed packages should not show it at all.

**Additional context**
#899

ankitsridhar16:
Closes #927

## 📑 Description
- Display status message in glasskube describe command
- Message should be shown below the status, but non-installed packages should not show it at all.

## ✅ Checks


cmd/glasskube/cmd/describe.go
```diff
@@ -145,6 +146,7 @@ var describeCmd = &cobra.Command{
 					fmt.Println(bold("    Namespace:  "), pkg.Namespace)
 					fmt.Println(bold("    Version:    "), version(&pkg, latestVersion))
 					fmt.Println(bold("    Status:     "), status(&pkg))
+					fmt.Println(bold("    Message:     "), message(&pkg))
```
christophenne, line 149: after "Message:" there is one whitespace too much.

cmd/glasskube/cmd/describe.go
```diff
@@ -136,6 +136,7 @@ var describeCmd = &cobra.Command{
 			if !pkg.IsNil() {
 				fmt.Println(bold("Version:    "), version(pkg, latestVersion))
 				fmt.Println(bold("Status:     "), status(pkg))
+				fmt.Println(bold("Message:     "), message(pkg))
```
christophenne, line 139: after "Message:" there is one whitespace too much.

cmd/glasskube/cmd/describe.go
```diff
@@ -329,6 +331,17 @@ func status(pkg ctrlpkg.Package) string {
 	}
 }
 
+func message(pkg ctrlpkg.Package) string {
+	pkgStatus := client.GetStatusOrPending(pkg)
```
christophenne, line 335: The `status` function is also getting the status by calling this method, so I think at this point it makes sense to get the `pkgStatus` in the calling function, and then pass the status object into `status` and `message`.

cmd/glasskube/cmd/describe.go
```diff
@@ -329,6 +331,17 @@ func status(pkg ctrlpkg.Package) string {
...
+	if pkgStatus != nil {
+		if pkgStatus.Status == string(condition.Ready) {
```
christophenne, line 337: I think for the message it's not necessary to print it in green. Also, the message should always be printed, not only when the status is ready.

christophenne (changes requested):
Thanks @ankitsridhar16, that's a good start, but I have left some comments about the implementation. Please have a look and let me know, if anything is unclear.

cmd/glasskube/cmd/describe.go
ankitsridhar16: can you elaborate a bit on the expected functionality here ?

cmd/glasskube/cmd/describe.go
ankitsridhar16: I have removed the check for ready condition and also removed color formatting. Can you take a look and review @christophenne

cmd/glasskube/cmd/describe.go
christophenne: `pkgStatus := client.GetStatusOrPending(pkg)` should be done before line 138, and then `pkgStatus` should be passed into `status()` and `message()`. This way it gets evaluated only once, and can be passed into the functions.

ankitsridhar16: @christophenne added new changes let me know

christophenne (approved):

PR #944, opened 2024-07-03
feat(cli): add `dry-run` option to configure command

### LINKED ISSUE #661: Add `--dry-run` support for `glasskube configure`
See #629 for details.

The `--dry-run` flag can be used by users, who want to simulate configuring a package, but don't actually want to configure it.

doncicuto:
Closes #661

## 📑 Description
This PR adds the `--dry-run` flag to the `glasskube configure` command

## ✅ Checks

## ℹ Additional Information

- A bug has been fixed that prevents the configure command to update the values. The `pkgClient.ClusterPackages().Get(ctx, pkg.Name, pkg)` call overwrote the values to be updated with the old values in the pkg so the update operation wasn't using the new values. The same happened with `pkgClient.Packages(pkg.Namespace).Get(ctx, pkg.Name, pkg)`. These calls have been removed as `getPackageOrClusterPackage` was already checking if the package was available to configure it anyway
- Several update functions now require and opts argument due to the required change in readWriteClientInterface
- As commented in the issue a new DryRunOptions type option has been added and replaced the dry-run flag definition in the install and bootstrap command

Here are some screenshots showing the configure command in action.

Initial values:

image

I try to update the prometheusStorageSize to 12Gi in dry run:

image

The value for prometheusStorageSize is still 8Gi:

image

I try to update the prometheusStorageSize to 12Gi without dry run:

image

The value for prometheusStorageSize has changed to 12Gi:

image


cmd/glasskube/cmd/configure.go
```diff
@@ -74,28 +83,20 @@ func runConfigure(cmd *cobra.Command, args []string) {
...
 
 	switch pkg := pkg.(type) {
 	case *v1alpha1.ClusterPackage:
-		if err := pkgClient.ClusterPackages().Get(ctx, pkg.Name, pkg); err != nil {
-			// Don't exit, we can still try to call update ...
-			fmt.Fprintf(os.Stderr, "⚠️  error fetching package: %v\n", err)
-		}
-
-		if err := pkgClient.ClusterPackages().Update(ctx, pkg); err != nil {
+		if err := pkgClient.ClusterPackages().Update(ctx, pkg, opts); err != nil {
```
kosmoz, line 94: Thank you for finding and fixing this bug, however the reason why we fetched the package again was that it might take the user some time to complete the configuration wizard, so the chance that the resource would have been updated in the meantime is relatively high. This leads to an error because k8s only allows updating on the most recent resource version. 

Hence, I think something like this would be a better fix:
```go
values := pkg.Spec.Values
if err := pkgClient.ClusterPackages().Get(ctx, pkg.Name, pkg); err != nil {
	// Don't exit, we can still try to call update ...
	fmt.Fprintf(os.Stderr, "⚠️  error fetching package: %v\n", err)
}
pkg.Spec.Values = values
```

kosmoz (changes requested):

cmd/glasskube/cmd/configure.go
doncicuto: I see @kosmoz thanks for the explanation, I'll proceed that way

cmd/glasskube/cmd/configure.go
```diff
@@ -74,28 +83,34 @@ func runConfigure(cmd *cobra.Command, args []string) {
...
-		cancel()
+	if !configureCmdOptions.DryRun {
+		if !cliutils.YesNoPrompt("Continue?", true) {
+			cancel()
+		}
 	}
 
 	switch pkg := pkg.(type) {
 	case *v1alpha1.ClusterPackage:
+		values := pkg.Spec.Values
```
kosmoz, line 94: There is another bug here: because maps are always pointers in go, `values` will also get overwritten by re-fetching the package. You can change it to `maps.Clone(pkg.Spec.Values)` then it should work.

kosmoz (changes requested):

cmd/glasskube/cmd/configure.go
doncicuto: I missed this, while I try to update the configuration it seemed to work fine though, thanks @kosmoz

kosmoz (approved):

PR #954, opened 2024-07-04
feat(package-operator): support for `resourceName`s in patches

christophenne:
## 📑 Description
Needed for https://github.com/glasskube/packages/pull/139

## ✅ Checks

### DIFF
--- a/internal/manifestvalues/patch.go
+++ b/internal/manifestvalues/patch.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/json"
+	"regexp"
 	"text/template"
 
 	jsonpatch "github.com/evanphx/json-patch/v5"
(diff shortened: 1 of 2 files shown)


internal/manifestvalues/patch.go
```diff
@@ -17,6 +18,17 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+var resourceNameExpr = regexp.MustCompile("[^a-zA-Z0-9]+")
+
+func resourceName(s string) string {
+	return resourceNameExpr.ReplaceAllString(s, "-")
+}
```
kosmoz, line 25: We already have something similar here: internal/names/utils.go would it make sense to use that instead? The difference would be of course, that we would get double dashes instead of single

christophenne: Oh nice, yes that should work, I will change it.

christophenne: With the existing `escapeResourceName` we do not get any dashes at all, because the points are not replaced. I am not sure if that is correct or not? 

If secret names are allowed to have '.' in them, we can use it (can't really tell from here https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)

christophenne:
As discussed, since '.' are allowed in resource names, there is no real need to support some additional resource name replacements. For the quickwit domain value (https://github.com/glasskube/packages/pull/139) we will simply use the kubernetes validation regex as a pattern constraint, and use the exact domain name as secret name.

PR #955, opened 2024-07-04
feat(ui): open and update buttons for namespaced packages

christophenne:
Ref #803 

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -501,7 +501,7 @@ func (s *server) packages(w http.ResponseWriter, r *http.Request) {
 
 	packageUpdateAvailable := map[string]bool{}
 	var installed []*list.PackagesWithStatus
-	var available []*list.PackagesWithStatus
+	var available []*repotypes.PackageRepoIndexItem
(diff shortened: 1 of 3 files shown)


christophenne:
Screenshot from 2024-07-05 11-11-10

I also made the rows hoverable, because I think already with three rows its "difficult" to make sure you press the button in the correct row, but I'm still unsure about it. 
Other than that, I think something is off in general, but I can't tell what. 

Open for any input.

kosmoz:
> Open for any input.

Have you tried `table-striped` as well? I think it might work too, although the two are not mutually exclusive, so we can also use both

christophenne:
> 
> Have you tried `table-striped` as well? I think it might work too, although the two are not mutually exclusive, so we can also use both

yes, but also not so sure about this: 

Screenshot from 2024-07-05 13-21-41

kosmoz:
I see, we can think about this a little more. For now, adding `table-hover` is definitely an improvement regardless :+1:

kosmoz (approved):

Issue #928, opened 2024-07-02
`glasskube ls --kind package` should also support `-o` flag


christophenne:
**Is your feature request related to a problem? Please describe.**
`-o` (output) flag is not yet implemented for listing packages, only for clusterpackages.

**Describe the solution you'd like**
Similar to `glasskube ls --kind clusterpackage` the command should also support `yaml` and `json` output for packages.

Arsh0023:
Hi @christophenne, can I take this issue ?

PR #967, opened 2024-07-07
fix(cli): `list` command with `--output` flag respects `--kind` flag

Arsh0023:
Ref #928 

## 📑 Description
Have added support for for -o , --output flag for kind package. Earlier this was only functional for kind clusterpackage.

## Preview
Screenshot from 2024-07-07 16-41-44

### DIFF
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -66,22 +66,22 @@ var listCmd = &cobra.Command{
 		}
 		noPkgs := len(pkgs) == 0 && listCmdOptions.Kind != KindClusterPackage
 		noClPkgs := len(clPkgs) == 0 && listCmdOptions.Kind != KindPackage
-		if noPkgs {
-			handleEmptyList("packages")
-		} else if len(pkgs) > 0 {
-			printPackageTable(pkgs)
-			if listCmdOptions.Kind != KindPackage {
-				fmt.Fprintln(os.Stderr, "")
+		if listCmdOptions.Output == OutputFormatJSON {
+			printPackageJSON(allPkgs(clPkgs, pkgs))
+		} else if listCmdOptions.Output == OutputFormatYAML {
+			printPackageYAML(allPkgs(clPkgs, pkgs))
+		} else {
+			if noPkgs {
+				handleEmptyList("packages")
+			} else if len(pkgs) > 0 {
+				printPackageTable(pkgs)
+				if listCmdOptions.Kind != KindPackage {
+					fmt.Fprintln(os.Stderr, "")
+				}
 			}
-		}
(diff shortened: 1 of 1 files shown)


kosmoz (changes requested):
The JSON output should always be a single list object. This implementation leads to printing multiple lists. To achieve this, you might have to change the logic of this command a bit:

(pseudo code)

```go
if output == json {
  printPackageJSON(allPkgs(clPkgs, pkgs))
} else if output == yaml {
  printPackageYAML(allPkgs(clPkgs, pkgs))
} else {
  if noPkgs {
    handleEmptyList("packages")
  } else {
    printPackageTable(pkgs)
  }
  if noClPkgs {
    handleEmptyList("clusterpackages")
  } else {
    printClusterPackageTable(clPkgs)
  }
}
```

something like that … the `allPkgs` function doesn't exist yet, it combines all clusterpackages and packages and returns a []ctrlpkg.Package.

Let me know if you have any questions

Arsh0023:
Okay thanks @kosmoz for the review, will make the required changes.

cmd/glasskube/cmd/list.go
```diff
@@ -133,6 +130,15 @@ func handleEmptyList(resource string) {
 	}
 }
 
+func allPkgs(clpkgs []*list.PackageWithStatus, pkgs []*list.PackagesWithStatus) []*list.PackageWithStatus {
+	result := make([]*list.PackageWithStatus, 0, len(clpkgs)+len(pkgs))
+	result = append(result, clpkgs...)
+	for _, pkg := range pkgs {
+		result = append(result, pkg.Packages...)
```
christophenne, line 137: The problem with this approach is, that when a package is not installed at all, it will not appear in the resulting list at all (because `pkg.Packages` is empty). On the other hand, if a certain clusterpackage is not installed, it will be part of the resulting list. So that would be inconsistent. 

I think you would need to check if `pkg.Packages` is empty here. If not, you can do what you are doing right now, but if it is empty, you would need to construct an "empty" `list.PackageWithStatus`, where you just set the `MetaIndexItem`.

```
if len(pkg.Packages) > 0 {
	result = append(result, pkg.Packages...)
} else {
	result = append(result, &list.PackageWithStatus{
		MetaIndexItem: pkg.MetaIndexItem,
	})
}
```

christophenne (changes requested):
Please also test the command with `--kind package` and `--kind clusterpackage` and check whether it is doing the same as before. It seems that right it's also printing the table header of the non-shown table, which should not be the case.

kosmoz (approved):

kosmoz (approved):

christophenne (approved):

Issue #948, opened 2024-07-04
Bootstrap Stuck on 96% (Applying glasskube (PackageRepository))


eefret:
**Describe the bug**
Cant get to bootstrap glasskube, I'm using an EKS cluster with Fargate nodes, tried purging and reinstalling.

**To Reproduce**
1. Create an EKS cluster
2. glasskube bootstrap

**Expected behavior**
Should complete bootstrap

**Screenshots**
![image](https://github.com/glasskube/glasskube/assets/863005/49551a12-9c93-4652-8b7b-ae0ba422f64a)

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: v1.28.9-eks-036c24b
 - Provider: AWS

**Glasskube Info (please complete the following information):**
 - Package Operator version: v0.12.1
 - CLI version: v0.12.1

**Additional context**
kubectl get nodes -o wide
![image](https://github.com/glasskube/glasskube/assets/863005/745edb8a-54b6-446d-a03e-5953d02739d7)

pmig:
Thanks for raising this issue. Glasskube currently internally uses the Flux `helm-controller`. This is not yet well documented (https://github.com/glasskube/glasskube/issues/917). In this case you would need to bootstrap Glasskube in the `--type=slim` version (https://glasskube.dev/docs/getting-started/bootstrap/).

Can this be also the case in your cluster?

kosmoz:
Hey @eefret, could you please let us know the output of `kubectl describe pods -n glasskube-system`? That might help us finding the issue here.

kosmoz:
Ah I see, that would explain the problem! 

Since I don't think this is something we can solve in the general sense, we should add a "Common issues" section to the docs and explain it there.

PR #971, opened 2024-07-08
fix(cli): improve bootstrap progressbar accuracy and error message

kosmoz:
Ref #948 

## 📑 Description
1. The progressbar would report an incorrect message most of the time because of the default throttle config
2. The error message in case a workload does not become ready was not helpful

## ✅ Checks

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -207,6 +207,7 @@ func (c *BootstrapClient) applyManifests(
 	bar := progressbar.Default(int64(len(objs)), "Applying manifests")
 	progressbar.OptionClearOnFinish()(bar)
 	progressbar.OptionOnCompletion(nil)(bar)
+	progressbar.OptionThrottle(0)(bar)
 	defer func(bar *progressbar.ProgressBar) { _ = bar.Exit() }(bar)
 
 	var checkWorkloads []*unstructured.Unstructured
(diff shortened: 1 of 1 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -253,7 +254,9 @@ func (c *BootstrapClient) applyManifests(
 
 	for _, obj := range checkWorkloads {
 		bar.Describe(fmt.Sprintf("Checking Status of %v (%v)", obj.GetName(), obj.GetKind()))
-		if err := c.checkWorkloadReady(obj.GetNamespace(), obj.GetName(), obj.GetKind(), 5*time.Minute); err != nil {
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+		defer cancel() // release the context in case checkWorkloadReady returned early
```
christophenne, line 258: `defer` inside a loop seems a bit weird to me, also Goland tells me a warning there: "Possible resource leak, 'defer' is called in the 'for' loop". Wouldn't the `cancel` function always be the one of the last loop iteration?

I think this could be made a bit clearer by either introducing a function that's called inside the loop and do the `defer` there, or by calling `cancel()` after `checkWorkloadReady` returns.

christophenne: ah okay, I guess it's not a problem after all: https://go.dev/tour/flowcontrol/13

christophenne (approved):

pkg/bootstrap/bootstrap.go
kosmoz: Huh, I actually thought that defer would make the call when the variable scope ends, so TIL. Would this fix the IDE warning?

```go
defer func() { cancel() }()
```

christophenne: Nope, also considered as a possible resource leak

kosmoz: should be fixed now

christophenne (approved):

PR #985, opened 2024-07-10
feat(cli): add --dry-run to update command

### LINKED ISSUE #660: Add `--dry-run` support for `glasskube update`
See #629 for details.

The `--dry-run` flag can be used by users, who want to simulate updating a package, but don't actually want to update it.

hanshal101:
Closes #660 
Fixes #660 

## 📑 Description

This PR adds the `--dry-run` tags to the update command.
## ✅ Checks

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -176,19 +176,19 @@ outer:
 	return &tx, nil
 }
 
-func (c *updater) Apply(ctx context.Context, tx *UpdateTransaction, DryRun bool) ([]ctrlpkg.Package, error) {
-	return c.apply(ctx, tx, false, DryRun)
+type ApplyUpdateOptions struct {
+	Blocking bool
+	DryRun   bool
 }
 
-func (c *updater) ApplyBlocking(ctx context.Context, tx *UpdateTransaction, DryRun bool) ([]ctrlpkg.Package, error) {
-	return c.apply(ctx, tx, true, DryRun)
+func (c *updater) Apply(ctx context.Context, tx *UpdateTransaction, opts ApplyUpdateOptions) ([]ctrlpkg.Package, error) {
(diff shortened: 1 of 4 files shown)


pkg/update/update.go
```diff
@@ -176,15 +176,20 @@ outer:
 	return &tx, nil
 }
 
-func (c *updater) Apply(ctx context.Context, tx *UpdateTransaction) ([]ctrlpkg.Package, error) {
-	return c.apply(ctx, tx, false)
+func (c *updater) Apply(ctx context.Context, tx *UpdateTransaction, DryRun bool) ([]ctrlpkg.Package, error) {
```
christophenne, line 179: Since we are now having two bool options (blocking and dryrun), I suggest the following refactoring: there should only be one `Apply` method instead of two (`Apply`, `ApplyBlocking`), and it gets an `ApplyUpdateOptions` object.

```
type ApplyUpdateOptions struct {
    Blocking bool
    DryRun bool 
}
```

Of course that also means that at any point where `Apply` or `ApplyBlocking` has been called so far, it needs to be changed to use `Apply` with the correct options object.

christophenne (commented):
I haven't tested it yet, but I have a comment about the code design, please let me know what you think.

hanshal101:
I hope this what you meant.

christophenne (changes requested):
Yes that's what I meant :+1:

hanshal101:
Made all the changes here, I have removed the `Apply` function and renamed the `apply` function to `Apply`. Hope this helps.

christophenne:
> Made all the changes here, I have removed the `Apply` function and renamed the `apply` function to `Apply`. Hope this helps.

It's missing registering the dry run options, you need to `updateCmdOptions.DryRunOptions.AddFlagsToCommand(updateCmd)` in `init()` (like also done in `install` command).

hanshal101:
Cool @christophenne made the updates 👍

christophenne (approved):
LGTM!

PR #987, opened 2024-07-10
feat(ui): status message on package details page

### LINKED ISSUE #898: UI: Display status condition message on package details page
If a package is in the state "installation failed" a user has no further information why the installation failed or if he should do something. Most users won't be able to look up the package status conditions.

On the package detail page there should a message with the latest status description

Linking to the Kubernetes Dashboard for example (but this makes not that much sense)

**Additional context**
Related cli issue:  https://github.com/glasskube/glasskube/issues/899

Aryakoste:
Closes #898 

## 📑 Description
Added status badge and alert box below header whenever the status goes failed to show the message.

## ✅ Checks

### DIFF
--- a/internal/web/templates/components/pkg-detail-header.html
+++ b/internal/web/templates/components/pkg-detail-header.html
@@ -114,22 +114,26 @@
             Auto-Update:
             <strong>{{ .AutoUpdate }}</strong>
           </span>
-          <span 
-          class="badge p-1 fw-normal {{ if eq .Status.Status "Failed" }}bg-danger text-white border border-1 border-danger{{ else }}bg-body-secondary text-primary-emphasis border-primary border border-1 p-1 fw-normal{{ end }}"
-          data-bs-toggle="popover"
-          data-bs-title="Status"
-          data-bs-content="{{ .Status.Reason }}"
-          data-bs-html="true"
-          data-bs-trigger="focus"
-          tabindex="0">
-          Status: <strong>{{ .Status.Status }}</strong>
-        </span>
+          <span
+            class="badge p-1 fw-normal {{ if eq .Status.Status " Failed" }}
+              bg-danger text-white border border-1 border-danger
+            {{ else }}
+              bg-body-secondary text-primary-emphasis border-primary border border-1 p-1 fw-normal
+            {{ end }}"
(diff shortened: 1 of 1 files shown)


internal/web/templates/components/pkg-detail-header.html
```diff
@@ -114,7 +114,26 @@ <h1 class="text-reset m-0">
             Auto-Update:
             <strong>{{ .AutoUpdate }}</strong>
           </span>
+          <span
+            class="badge p-1 fw-normal {{ if eq .Status.Status " Failed" }}
+              bg-danger text-white border border-1 border-danger
+            {{ else }}
+              bg-body-secondary text-primary-emphasis border-primary border border-1 p-1 fw-normal
+            {{ end }}"
+            data-bs-toggle="popover"
```
christophenne, line 123: I think all these `data-bs-` attributes can be removed now that we don't want any tooltip/popover, right?

internal/web/templates/components/pkg-detail-header.html
```diff
@@ -114,7 +114,26 @@ <h1 class="text-reset m-0">
             Auto-Update:
             <strong>{{ .AutoUpdate }}</strong>
           </span>
+          <span
+            class="badge p-1 fw-normal {{ if eq .Status.Status " Failed" }}
```
christophenne, line 118: `" Failed"` should be `"Failed"`, but to be honest now that I have tried it out, I don't know about this whole `if/else`.
I think the failed status could be shown the same as every other status without any special formatting? Because there is already the flashy red "Installation Failed" on the top right + the alert box below to show the message. But I'm not sure.

christophenne: If you want to keep it that way, I suggest to change the template code a bit, since right now it's confusing to read but also seems to mess with IDEs (at least mine). As an alternative, you define make two little subtemplates in the same file, one for regular status badge, one for failed status badge. Then it would be a bit easier to read, then the `if / else` templating inside the `class` attribute.

christophenne (changes requested):
Thanks a lot, looks pretty good to me so far :) only some comments/suggestions.

christophenne (approved):

Issue #789, opened 2024-06-10
UI breaks when several tabs are opened


christophenne:
**Describe the bug**
Open the UI in 10 tabs, the UI becomes completely unresponsive. 

**To Reproduce**

**Expected behavior**
UI works. 

**Screenshots**

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: 1.30
 - Provider: minikube

**Glasskube Info (please complete the following information):**
 - Package Operator version: dev
 - CLI version: dev
 
**Additional context**

christophenne:
The UI doesn't load because the browser is stalling the request. This happens, when too many connections to the server are opened already. See https://stackoverflow.com/a/29564247 for more details. In Chrome it's deterministically reproducible when opening the 7th tab. Also, if you close one of the opened localhost:8580-tabs, a connection becomes available, and the stalling tab will load then. The 6 already opened tabs will continue to work. 

The limit of 6 is only for HTTP/1.1, so a workaround would be to make the server HTTP-2 ready, this could serve as a reference: https://posener.github.io/http2/ (we need to serve with https then obviously).

@pmig FYI

pmig:
Let's go for http2 🚀

christophenne:
@pmig @kosmoz I had another thought/idea about this. I think this is more likely to happen because browsers seem to be more aggressive when trying to reconnect to an SSE endpoint. There is an exponential backoff to reconnect in the htmx extension, but it seems this is just a fallback (https://github.com/bigskysoftware/htmx-extensions/blob/main/src/sse/README.md#automatic-reconnection) – firefox and chrome seem to be reconnecting every 1 - 2 seconds until forever. 

I'm assuming that it's more likely for package maintainers to run into this issue, because they will stop and restart the server more often during integrating a package (also "hopefully" because of our current hardcoded 5 minutes cache (#756)). If they don't close old tabs, these tabs will continue to try to reconnect and succeed at that because of the small retry interval. 

I think a pragmatic fix would be to stop the reconnect all together, because it doesn't actually bring any benefit to the user at all. First of all, when the connection is lost, we show the "disconnected" banner. And when it is actually reconnected, we show the "you might be out of date, please refresh" message instead. So we don't really support reconnecting and re-syncing the client to be on the latest state anway.
That would mean that the user sees the "disconnected" banner until they fully refresh the page, otherwise the page is simply outdated, which will be fine for "old" tabs. 

[...]

PR #991, opened 2024-07-12
fix(ui): gracefully close sse and refactor to toast feedback

christophenne:
Ref #789 

## 📑 Description
In this PR we introduce a more graceful shutdown of the server. On shutdown an `close` event is broadcasted, which is picked up by the htmx sse extension to close the `EventSource` in the browser, therefore avoiding endless reconnection attempts. This should somewhat decrease the problems when restarting the UI very often and the browser stalling the requests when a few tabs have been opened (as described in #789). 

Apart from that, error handling has been refactored to show toast elements that are always on screen. 

## ✅ Checks


internal/web/server.go
```diff
@@ -256,15 +264,37 @@ func (s *server) Start(ctx context.Context) error {
...
+
+	<-s.httpServerHasShutdownCh
+	cliutils.ExitFromSignal(receivedSig)
+
 	return nil
 }
 
+func (s *server) shutdown() {
+	close(s.stopCh)
+	if err := s.httpServer.Shutdown(context.Background()); err != nil {
```
kosmoz, line 292: Since shutdown may wait indefinitely until all connections are closed, maybe it would be a good idea to pass a context with timeout here? Something around 10s?

christophenne: Good idea :+1:

kosmoz (approved):

PR #1021, opened 2024-07-21
chore(logging): extend serve command with klog

### LINKED ISSUE #836: Extend logging that happens during `serve` command
Currently the `glasskube serve` command only logs if a user installs a package. So only following the logs it seems that the installation will not be finished.

All (or most) status changes should be logged to `Stderr`. This can be noisy as this also might help us to reproduce and fix bugs if users can copy past us the ui logs. 

**Additional context**
This also might help investigating the "out-of-sync" task (#838).

oversampling:
Closes #836 

## 📑 Description
Currently serve command only print out the error log, instead this PR going to use Klog to logging out the error. This PR going to use structured logging with content of error msg, context msg and error type. 

PS: If there is change required for the log content, welcome to let me know. 

## ✅ Checks

## ℹ Additional Information
_Before_
image
_After_
image

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -4470,17 +4470,17 @@
       "integrity": "sha512-I4q9QU9MQv4oEOz4tAHJtNz1cwuLxn2F3xcc2iV5WdqLPpUnj30aUuxt1mAxYTG+oe8CZMV/+6rU4S4gRDzqtQ=="
     },
     "node_modules/@typescript-eslint/eslint-plugin": {
-      "version": "7.17.0",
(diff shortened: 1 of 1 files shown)


christophenne:
Hi @oversampling I forgot this when we last talked, but we also want to log all observed changes of our custom resources. You can implement this in the `init*StoreAndController` functions in `server.go` – here we react to the observed changes (see `AddFunc`, `UpdateFunc`, `DeleteFunc`). There should be a log output for every of these changes, probably on a debug level? I am not sure yet, but it would be nice if you could prepare the PR in that direction and then we will have a look at the log level. Thanks!

PR #1026, opened 2024-07-22
feat(cli): add `glasskube configure` showing an error for package with no value definitions

### LINKED ISSUE #867: `glasskube configure` should show special message when there are no values to configure
When a package has no configuration values, the `glasskube configure` command should show "This package has no configuration values" and quit. 

Example: `glasskube configure cert-manager`.

utsab818:
Closes #867 

## 📑 Description
`glasskube configure <package-without-configuration-values>`  will throw message as 'This package has no configuration values'
and exit.

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/configure.go
+++ b/cmd/glasskube/cmd/configure.go
@@ -59,11 +59,6 @@ func runConfigure(cmd *cobra.Command, args []string) {
 		cliutils.ExitWithError()
 	}
 
-	if len(pkg.GetSpec().Values) == 0 {
-		fmt.Fprintln(os.Stderr, "❌ This package has no configuration values")
-		cliutils.ExitWithError()
-	}
-
 	if configureCmdOptions.IsValuesSet() {
 		if values, err := configureCmdOptions.ParseValues(pkg.GetSpec().Values); err != nil {
 			fmt.Fprintf(os.Stderr, "❌ invalid values in command line flags: %v\n", err)
@@ -77,6 +72,9 @@ func runConfigure(cmd *cobra.Command, args []string) {
 			cliutils.ExitWithError()
 		} else if values, err := cli.Configure(*pkgManifest, pkg.GetSpec().Values); err != nil {
 			fmt.Fprintf(os.Stderr, "❌ error during configure: %v\n", err)
+			cliutils.ExitWithError()
+		} else if len(pkgManifest.ValueDefinitions) == 0 {
+			fmt.Fprintln(os.Stderr, "❌ this package has no configuration values:")
 			cliutils.ExitWithError()
 		} else {
 			pkg.GetSpec().Values = values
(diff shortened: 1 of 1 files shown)


kosmoz (changes requested):
Hi @utsab818, thanks for your PR!

This implementation does not work for all edge cases. For example, there might be a package where there are value definitions in the manifest, but no value configurations on the installed package. 

The solution is therefore to check `pkgManifest.ValueDefinitions` instead.

utsab818:
> Hi @utsab818, thanks for your PR!
> 
> This implementation does not work for all edge cases. For example, there might be a package where there are value definitions in the manifest, but no value configurations on the installed package.
> 
> The solution is therefore to check `pkgManifest.ValueDefinitions` instead.

cmd/glasskube/cmd/configure.go
```diff
@@ -73,6 +73,9 @@ func runConfigure(cmd *cobra.Command, args []string) {
 		} else if values, err := cli.Configure(*pkgManifest, pkg.GetSpec().Values); err != nil {
 			fmt.Fprintf(os.Stderr, "❌ error during configure: %v\n", err)
 			cliutils.ExitWithError()
+		} else if len(pkgManifest.ValueDefinitions) == 0 {
+			fmt.Fprintln(os.Stderr, "❌ this package has no configuration values:")
```
kosmoz, line 77:
```suggestion
			fmt.Fprintln(os.Stderr, "❌ this package has no configuration values")
```

kosmoz (changes requested):

cmd/glasskube/cmd/configure.go
```diff
@@ -73,6 +73,9 @@ func runConfigure(cmd *cobra.Command, args []string) {
 		} else if values, err := cli.Configure(*pkgManifest, pkg.GetSpec().Values); err != nil {
 			fmt.Fprintf(os.Stderr, "❌ error during configure: %v\n", err)
 			cliutils.ExitWithError()
+		} else if len(pkgManifest.ValueDefinitions) == 0 {
+			fmt.Fprintln(os.Stderr, "❌ this package has no configuration values")
+			cliutils.ExitWithError()
```
kosmoz, line 78: checking before calling configure makes more sense

```suggestion
		} else if len(pkgManifest.ValueDefinitions) == 0 {
			fmt.Fprintln(os.Stderr, "❌ this package has no configuration values")
			cliutils.ExitWithError()
		} else if values, err := cli.Configure(*pkgManifest, pkg.GetSpec().Values); err != nil {
			fmt.Fprintf(os.Stderr, "❌ error during configure: %v\n", err)
			cliutils.ExitWithError()
```

kosmoz (approved):

utsab818:
@kosmoz Thankyou for your guidance.

PR #1027, opened 2024-07-22
feat(cli): add `--host` flag for `serve` to change the hostname for the listener

### LINKED ISSUE #993: Listen on other NIC instead of local host for `glasskube  serve`
I run glasskube on a remote server without UI. the default serve command listen on 127.0.0.1 only. I check the help message and could find how to change it to 0.0.0.0.

```
./glasskube  serve --help
Start server and open the UI.

Usage:
  glasskube serve [flags]

Aliases:
  serve, start, ui

Flags:
  -h, --help            help for serve
  -l, --log-level int   Level for additional logging, where 0 is the least verbose
  -p, --port int        Port for the webserver (default 8580)
  -s, --skip-open       Skip opening the browser

Global Flags:
      --kubeconfig string   Path to the kubeconfig file, whose current-context will be used (defaults to xxx.kube/config)
      --no-progress         Prevent progress logging to the cli
      --non-interactive     Run in non-interactive mode. If interactivity would be required, the command will terminate with a non-zero exit code.
      --skip-update-check   Do not check for Glasskube updates

```

It should be able to listen on other address besdies localhost.

kosmoz:
Closes #993  

## 📑 Description
Adds a new flag to the `serve` command: `--host`

I also took the opportunity and cleaned up the "use different port" feature a little. 

## ✅ Checks

## ℹ Additional Information

I discovered a potential bug related to the cache verifier: While the cli is waiting for user input after the question "Should glasskube use a different port?", the cache verifier already starts printing warnings and eventually the command is stopped. This behavior already exists in current `main`. To reproduce (works 100%): Run `glasskube serve` in two separate terminals.

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -79,7 +79,7 @@ func init() {
 
 type ServerOptions struct {
 	Host               string
-	Port               int32
+	Port               string
 	Kubeconfig         string
 	LogLevel           int
 	SkipOpeningBrowser bool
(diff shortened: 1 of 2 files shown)


internal/web/server.go
```diff
@@ -227,33 +227,28 @@ func (s *server) Start(ctx context.Context) error {
...
-	s.listener, err = net.Listen("tcp", bindAddr)
+	s.listener, err = net.Listen("tcp", net.JoinHostPort(s.Host, s.Port))
 	if err != nil {
-		// Checks if Port Conflict Error exists
-		if isPortConflictError(err) {
-			userInput := cliutils.YesNoPrompt(
-				"Port is already in use.\nShould glasskube use a different port? (Y/n): ", true)
-			if userInput {
-				s.listener, err = net.Listen("tcp", ":0")
+		// if the error is "address already in use", try to get the OS to assign a random free port
```
christophenne, line 232: I think this error handling should be in the command rather than the server itself. That could make the server start method a bit more straightforward, and the `cliutils.YesNoPrompt` would actually only be called on the CLI layer. But I would say that's something for another day.

christophenne (approved):
> I discovered a potential bug related to the cache verifier: While the cli is waiting for user input after the question "Should glasskube use a different port?", the cache verifier already starts printing warnings and eventually the command is stopped. This behavior already exists in current main. To reproduce (works 100%): Run glasskube serve in two separate terminals.

Interesting, don't really understand it yet but I'll take a look.

internal/web/server.go
kosmoz: That's a fair point, but I think it might come with some more changes, since the server would have to ensure that all watchers, timers, etc are closed in case of an error (maybe related to the potential bug described above). Ideally, the server would perform any init logic between creating the listener (thus knowing that the address/port is "claimed") and starting the http server. Also, any teardown logic should probably be moved to a deferred function.

christophenne:
> To reproduce (works 100%): Run glasskube serve in two separate terminals.

Not 100%, only when there is more than one clusterpackage installed apparently :sweat_smile: . If I have exactly one (`cert-manager`) it works fine. I will debug this and try to find out why the cached store only holds one package, this could be a good lead to solve the out of sync issues too, because its actually reproducible. And I'm 99% sure it has to do with the informer/cache stuff. 
Anyway, for now it's okay that the cache verifier and all the background operations are running at this point. Whether or not the local port can be used has no influence on setting up the kubernetes clients, watches, and eventually the cache verifier.

PR #1030, opened 2024-07-22
feat(cli): added `--repositroy` flag for `glasskube list` command to filter based on repo

### LINKED ISSUE #750: `--repository` flag for `glasskube list` command
`glasskube ls` always shows the packages of all configured repositories, but there should be a way to look at packages of one specific repository.

`glasskube ls` should have a flag `--repository`, which can be used this way: `glasskube ls --repository [repoName]`. Only the packages defined in the given repo will then be listed. 

**Additional context**

Srikrishnabh:
Closes #750 

## 📑 Description

added new flag `--repository` to `glasskube list` command to filter based on repo

## ✅ Checks


kosmoz (changes requested):
If we know that we only want to fetch one repo, it would be much more efficient to use a default RepoClient instead of the RepoMetaClient. For this, just change `l.repoClient.Meta()` to `l.repoClient.ForRepoWithName(options.Repository)` if a repository name is present (line 153). I think that should completely remove the necessity to do any filtering later but you will have to turn the `PackageIndex` into a `PackageMetaIndex`.

Srikrishnabh:
> If we know that we only want to fetch one repo, it would be much more efficient to use a default RepoClient instead of the RepoMetaClient. For this, just change `l.repoClient.Meta()` to `l.repoClient.ForRepoWithName(options.Repository)` if a repository name is present (line 153). I think that should completely remove the necessity to do any filtering later but you will have to turn the `PackageIndex` into a `PackageMetaIndex`.

@kosmoz understood, I initially thought to use the `l.repoClient.ForRepoWithName(options.Repository)` but we have logic to set the installed version to the hightest semver across all repos. [line](https://github.com/glasskube/glasskube/blob/main/internal/repo/client/metaclient.go#L45). 
Is it fine if we return only current version for given repo?

kosmoz:
> > If we know that we only want to fetch one repo, it would be much more efficient to use a default RepoClient instead of the RepoMetaClient. For this, just change `l.repoClient.Meta()` to `l.repoClient.ForRepoWithName(options.Repository)` if a repository name is present (line 153). I think that should completely remove the necessity to do any filtering later but you will have to turn the `PackageIndex` into a `PackageMetaIndex`.
> 
> @kosmoz understood, I initially thought to use the `l.repoClient.ForRepoWithName(options.Repository)` but we have logic to set the installed version to the hightest semver across all repos. [line](https://github.com/glasskube/glasskube/blob/main/internal/repo/client/metaclient.go#L45). Is it fine if we return only current version for given repo?

Yes, I think that would be the expected behavior.

Srikrishnabh:
@kosmoz can you please review.

internal/repo/client/metaclient.go
```diff
@@ -65,6 +66,27 @@ func (d metaclient) FetchMetaIndex(target *types.MetaIndex) error {
 	}
 }
 
+func (d metaclient) FetchMetaIndexForRepo(repo string, target *types.MetaIndex) error {
```
kosmoz, line 69: The logic here seems sound, but I think it's confusing to have a function for a specific repo on the metaclient. Since this feature is quite specific to the list package, I think that's where it should be kept. Can you please move this function to `pkg/list`, such that it can be called like this (you can choose a different name if you want):

```go
var index types.PackageRepoIndex
d.clientset.ForRepoWithName(repo).FetchPackageRepoIndex(&index)
metaIndex := indexToMetaIndex(index)
```

kosmoz (changes requested):
Please check the comment below. Apart from that, the feature works as expected.

PS: Please make sure to resolve any merge conflicts on your branch

internal/repo/client/metaclient.go
Srikrishnabh: makes sense, I have modified as suggested.

pkg/list/lister.go
```diff
@@ -206,9 +208,9 @@ func (l *lister) fetchRepoAndInstalled(ctx context.Context, options ListOptions,
 	// TODO what if a package is namespaced in one repository, and with the same name cluster scoped in another??
 
 	resultLs := make([]result, 0)
-	for i, indexPackage := range index.Packages {
+	for _, indexPackage := range index.Packages {
 		res := result{
-			IndexItem: &index.Packages[i],
+			IndexItem: &indexPackage,
```
kosmoz, line 213: I think this change should be reverted. 

Even though pointing to the loop variable is not an error anymore in Go 1.22, I think it's still good practice and easier to understand to point to the array element.

internal/repo/client/metaclient.go
```diff
@@ -4,12 +4,13 @@ import (
 	"context"
 	"slices"
 
+	"go.uber.org/multierr"
+
```
kosmoz, line 8: Please revert this

pkg/list/lister.go
```diff
@@ -242,3 +244,40 @@ func setPackageInfo(packageInfos v1alpha1.PackageInfoList, res *result, pkg ctrl
...
+		if err := l.repoClient.Meta().FetchMetaIndex(index); err != nil {
+			return fmt.Errorf("could not fetch package repository index: %w", err)
+		}
+		return nil
+	}
+	// fetch index for given repo
+	var repoIndex repotypes.PackageRepoIndex
+	err := l.repoClient.ForRepoWithName(options.Repository).FetchPackageRepoIndex(&repoIndex)
+	if err != nil {
+		return fmt.Errorf("could not fetch package repository index for repo %s : %w",
```
kosmoz, line 259:
```suggestion
		return fmt.Errorf("could not fetch package repository index for repo %s: %w",
```

kosmoz (changes requested):
Please have a look at my comments below. I think they should be minor changes. Afterwards, I think we are ready to merge :slightly_smiling_face:

kosmoz:
Hi @Srikrishnabh, how are you doing? Do you need any help with this PR?

Srikrishnabh:
@kosmoz Thanks for asking. Sorry for delay, I have addressed the review comments.

kosmoz (approved):

PR #1036, opened 2024-07-23
feat(cli): add support for filtering package list by name and namespace

### LINKED ISSUE #865: CLI: Add support for listing all packages for a package name
### Is your feature request related to a problem? Please describe.
`glasskube list` shows the list of all ClusterPackages or Packages or both. It should also be possible to list all packages with a particular `.Spec.PackageInfo.Name`.

### Describe the solution you'd like
When `glasskube list` is called with an argument, only the `Package`s with that `.Spec.PackageInfo.Name` should be listed. 

#### Examples

```sh
# shows a list of all Package resources with .Spec.PackageInfo.Name == "quickwit"
glasskube list quickwit
```
```sh
# shows a list of all Package resources in namespace "foo" with .Spec.PackageInfo.Name == "quickwit"
glasskube list quickwit --namespace foo
```
```sh
# shows an error message and call ExitWithError
glasskube list quickwit --kind clusterpackage
```

#### More details
- This operational mode is defined by `len(args) > 0`
- This mode implies that only namespace-scoped `Package`s should be shown, no `ClusterPackage`s (like `--kind package`). 
- In this mode, using the `--kind` flag is an error

yeahservice:
Closes #865

## 📑 Description
Adds the optional argument `<packageName>` and optional flag `--namespace`to the `list` cmd.

```
# shows a list of all Package resources with .Spec.PackageInfo.Name == "quickwit"
glasskube list quickwit
```
```
# shows a list of all Package resources in namespace "foo" with .Spec.PackageInfo.Name == "quickwit"
glasskube list quickwit --namespace foo
```
```
# shows an error message and calls ExitWithError
glasskube list --kind clusterpackage --namespace default
```
```
# shows an error message and calls ExitWithError
glasskube list quickwit --kind clusterpackage
```

## ✅ Checks

## ℹ Additional Information
* The empty list error handling has been replaced with just one single message `No <resource> found.` for every case. Due to the different options to combine flags, sometimes multiple error messages are valid. It is hard to decide which one takes precedence. A simple single error message is less confusing instead.
* Introducing tests for list command. There is some complexity on how the flags interact with each other. Introducing new flags needs changes on retrieval logic and output logic. Some of these locations can be easily overlooked when not tested. Nevertheless, this requires some refactoring to make commands testable with injection of dependencies. So these changes need to be discussed first and are therefore not yet fully implemented.
  * Commands are created with a constructor, isolating them for testing and injecting necessary mock dependencies
    * e.g. fake k8s client (not yet fully implemented)
  * Commands need to return errors instead of exiting directly. Exiting logic could be handled in main calling the root command.

I would request some feedback if tests and changes like this are desired. Since the refactoring takes some effort.

### DIFF
--- a/cmd/glasskube/cmd/list.go
+++ b/cmd/glasskube/cmd/list.go
@@ -23,14 +23,18 @@ type ListCmdOptions struct {
 	ShowLatestVersion bool
 	ShowMessage       bool
 	More              bool
+	packageName       string
 	OutputOptions
 	KindOptions
+	NamespaceOptions
 }
 
 func (o ListCmdOptions) toListOptions() list.ListOptions {
 	return list.ListOptions{
 		OnlyInstalled: o.ListInstalledOnly,
 		OnlyOutdated:  o.ListOutdatedOnly,
+		PackageName:   o.packageName,
+		Namespace:     o.Namespace,
 	}
 }
 
@@ -39,12 +43,13 @@ var listCmdOptions = ListCmdOptions{
 }
 
 var listCmd = &cobra.Command{
-	Use:     "list",
+	Use:     "list [<package-name>]",
 	Aliases: []string{"ls", "l"},
 	Short:   "List packages",
 	Long: "List packages. By default, all available packages of the given repository are shown, " +
 		"as well as their installation status in your cluster.\nYou can choose to only show installed packages.",
 	PreRun: cliutils.SetupClientContext(true, &rootCmdOptions.SkipUpdateCheck),
+	Args:   cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := cmd.Context()
 		if listCmdOptions.More {
@@ -52,11 +57,20 @@ var listCmd = &cobra.Command{
 			listCmdOptions.ShowDescription = true
 			listCmdOptions.ShowMessage = true
 		}
+		if len(args) > 0 {
+			listCmdOptions.packageName = args[0]
+		}
+		if listCmdOptions.Kind == KindClusterPackage &&
+			(listCmdOptions.packageName != "" || listCmdOptions.Namespace != "") {
+			fmt.Fprintf(os.Stderr, "Argument [<package-name>] or flag [--namespace] not supported with kind %s.\n",
+				KindClusterPackage)
+			cliutils.ExitWithError()
+		}
 		lister := list.NewListerWithRepoCache(ctx)
 		var clPkgs []*list.PackageWithStatus
 		var pkgs []*list.PackagesWithStatus
 		var err error
-		if listCmdOptions.Kind != KindPackage {
+		if listCmdOptions.Kind != KindPackage && listCmdOptions.packageName == "" && listCmdOptions.Namespace == "" {
 			clPkgs, err = lister.GetClusterPackagesWithStatus(ctx, listCmdOptions.toListOptions())
 			handleListErr(len(clPkgs), err, "clusterpackages")
 		}
@@ -65,7 +79,8 @@ var listCmd = &cobra.Command{
 			handleListErr(len(pkgs), err, "packages")
 		}
 		noPkgs := len(pkgs) == 0 && listCmdOptions.Kind != KindClusterPackage
-		noClPkgs := len(clPkgs) == 0 && listCmdOptions.Kind != KindPackage
+		noClPkgs := len(clPkgs) == 0 && listCmdOptions.Kind != KindPackage &&
+			listCmdOptions.packageName == "" && listCmdOptions.Namespace == ""
 		if listCmdOptions.Output == OutputFormatJSON {
 			printPackageJSON(allPkgs(clPkgs, pkgs))
 		} else if listCmdOptions.Output == OutputFormatYAML {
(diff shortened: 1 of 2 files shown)


cmd/glasskube/cmd/list.go
```diff
@@ -23,72 +23,88 @@ type ListCmdOptions struct {
...
-				printPackageYAML(clPkgs)
-			} else {
-				printClusterPackageTable(clPkgs)
+			if len(args) > 0 {
+				listCmdOptions.packageName = args[0]
+			}
+			if listCmdOptions.Kind == KindClusterPackage &&
+				(listCmdOptions.packageName != "" || listCmdOptions.Namespace != "") {
+				//TODO returning an error in RunE will also print cmd help; Desired or not?
+				return fmt.Errorf("argument or flag not supported with kind %s", KindClusterPackage)
```
kosmoz, line 67: We prefer `Run` over `RunE` for this reason. Logging the error and calling `cliutils.ExitWithError()` instead should be fine.

cmd/glasskube/cmd/list.go
```diff
@@ -23,72 +23,88 @@ type ListCmdOptions struct {
...
-		noClPkgs := len(clPkgs) == 0 && listCmdOptions.Kind != KindPackage
-		if noPkgs {
-			handleEmptyList("packages")
-		} else if len(pkgs) > 0 {
-			printPackageTable(pkgs)
-			if listCmdOptions.Kind != KindPackage {
-				fmt.Fprintln(os.Stderr, "")
+func newListCmd() *cobra.Command {
+	listCmd := &cobra.Command{
+		Use:     "list",
```
kosmoz, line 47:
```suggestion
		Use:     "list [<package-name>]",
```

kosmoz (commented):
Hi @yeahservice, thanks for your PR! I already have some preliminary feedback, please see below. Regarding your questions:

> * The empty list error handling has been replaced with just one single message `No <resource> found.` for every case. Due to the different options to combine flags, sometimes multiple error messages are valid. It is hard to decide which one takes precedence. A simple single error message is less confusing instead.

Sounds good to me. Any objections, @christophenne?

> * Introducing tests for list command. There is some complexity on how the flags interact with each other. Introducing new flags needs changes on retrieval logic and output logic. Some of these locations can be easily overlooked when not tested. Nevertheless, this requires some refactoring to make commands testable with injection of dependencies. So these changes need to be discussed first and are therefore not yet fully implemented.
>   
>   * Commands are created with a constructor, isolating them for testing and injecting necessary mock dependencies
>     
>     * e.g. fake k8s client (not yet fully implemented)
>   * Commands need to return errors instead of exiting directly. Exiting logic could be handled in main calling the root command.

I think more tests is almost never a bad thing. :smile: But I think there might be some things we could improve about the specific approach. 

For testing argument/flag handling, the validation could be extracted into a method on the `ListCmdOptions` type and could then be tested independently of the `Command`. If that's the only thing we want to test, I think this should be the way.

Anything that requires a bigger refactoring can be considered, but I would personally prefer having a separate issue/PR for that. If you are interested in doing that, you can create an issue and I will assign it to you right away :slightly_smiling_face:

yeahservice:
@kosmoz I have made the changes, please have another look if it fits the requirements.

I have removed the tests and the necessary refactoring. I agree with the separate PR. Although I sees some benefit in testing the whole listing command, since introducing new flags or arguments requires changes in 3~4 different functions. It would also cover well for regression.

Unit testing the single functions would also work, but requires more effort to get the same coverage. In case I have time, I will open an issue for contribution though!

pkg/list/lister.go
```diff
@@ -101,7 +103,8 @@ func (l *lister) GetPackagesWithStatus(
 			}
 		}
 		hasIncludedItems := len(ls) > 0
-		if hasIncludedItems || (!options.OnlyInstalled && !options.OnlyOutdated) {
+		if hasIncludedItems ||
+			(!options.OnlyInstalled && !options.OnlyOutdated && options.Namespace == "" && options.PackageName == "") {
```
kosmoz, line 107: IMO, `glasskube list quickwit` should still show the "not installed" line for that package if applicable, but with this change the output is "no package found" instead. 

I think checking the namespace makes sense but the "package name" part should be removed.

```suggestion
		if hasIncludedItems ||
			(!options.OnlyInstalled && !options.OnlyOutdated && options.Namespace == "") {
```

pkg/list/lister.go
```diff
@@ -118,10 +121,16 @@ func clusterPackageShouldBeIncluded(item *result, options ListOptions) bool {
...
+		if options.PackageName != "" && pkg.Spec.PackageInfo.Name != options.PackageName {
+			return false
 		}
-		return item.IndexItem != nil && pkg.Spec.PackageInfo.Version != item.IndexItem.LatestVersion
+		if options.OnlyOutdated && (item.IndexItem != nil && pkg.Spec.PackageInfo.Version == item.IndexItem.LatestVersion) {
+			return false
+		}
+		if options.Namespace != "" && pkg.Namespace != options.Namespace {
+			return false
+		}
```
kosmoz, line 132: We can do this much earlier by adding the namespace to the packages request (see https://github.com/yeahservice/glasskube/blob/0e224f90f8e8e27cd23cabb13abedd3d3440cef9/pkg/list/lister.go#L184)

kosmoz (changes requested):
This looks pretty good already, please have a look at the comments below. Also, please make sure to change your PR status from "draft" to "ready for review" if you think it's ready :slightly_smiling_face:

kosmoz (commented):
Hey @yeahservice, please have a look at my previous review, I think the comments are still unresolved. Let us know if you disagree with anything, I'm happy to discuss.

pkg/list/lister.go
yeahservice: The "not installed" packages are now also shown if the package name matches. All "not installed" packages are shown if no package name is specified.

kosmoz (approved):

PR #1050, opened 2024-07-30
feat: improve gitops support

christophenne:
Closes # 

## ✅ Checks


pmig (changes requested):
I think I came up with the best sub command naming:

`glasskube bootstrap git --url ...` :-)

cmd/glasskube/cmd/bootstrap_gitops.go
```diff
@@ -0,0 +1,126 @@
...
+		} else if installedVersion, err = semver.NewVersion(installedVersionRaw); err != nil {
+			fmt.Fprintf(os.Stderr, "could not parse installed version: %v\n", err)
+			cliutils.ExitWithError()
+		} else if installedVersion != nil {
+			fmt.Fprintf(os.Stderr, "gitops bootstrapping on a bootstrapped cluster is not supported yet. \n")
+			cliutils.ExitWithError()
+		}
+
+		currentContext := color.New(color.Bold).Sprint(clicontext.RawConfigFromContext(ctx).CurrentContext)
+		fmt.Fprintf(os.Stderr, "Glasskube and ArgoCD will be installed in context %s.\n", currentContext)
```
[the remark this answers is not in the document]
christophenne: What if we change "context" to "cluster"? We also don't talk about the `glasskube-system` namespace in any of the other commands, so I don't see a reason why that would be done here.

cmd/glasskube/cmd/bootstrap_gitops.go
```diff
@@ -0,0 +1,126 @@
...
+		// wait for argo to be ready, then print...
+		// TODO
+
+		fmt.Fprintf(os.Stderr, "DONE\n")
+
+	},
+}
+
+func init() {
+	bootstrapGitopsCmd.Flags().StringVar(&bootstrapGitopsCmdOptions.repo, "repo", "", "URL of the GitOps Repository")
```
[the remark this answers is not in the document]
christophenne: Not sure about the conventions, but since `--url` is required anyway, wouldn't it make sense that it is an argument instead? `glasskube bootstrap git https://...`

cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -0,0 +1,148 @@
...
+		} else if client, err := dynamic.NewForConfig(cfg); err != nil {
+			fmt.Fprintf(os.Stderr, "\nAn error occurred initializing the dynamic client:\n%v\n", err)
+			cliutils.ExitWithError()
+		} else {
+			for _, obj := range objs {
+				if _, err = client.Resource(schema.GroupVersionResource{
+					Group:    "argoproj.io",
+					Version:  "v1alpha1",
+					Resource: "applications",
+				}).Namespace(obj.GetNamespace()).Apply(ctx, obj.GetName(), &obj, metav1.ApplyOptions{
```
kosmoz, line 126: using a RESTMapper would be more robust, because it ensures that the resource can be handled by the cluster and would support different kinds of resource. We do the same in pkg/bootstrap

cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		fmt.Fprintf(os.Stderr, "\nglasskube argo-cd application applied successfully!\n\n")
+		fmt.Fprintf(os.Stderr, "You have successfully installed Glasskube and ArgoCD in this cluster.\n")
+		fmt.Fprintf(os.Stderr, "Right now, ArgoCD is starting up and will soon sync with your GitOps repo. ")
+		fmt.Fprintf(os.Stderr, "Please note, that this might take a couple of minutes!\n\n")
+		fmt.Fprintf(os.Stderr, "Run glasskube serve to open the Glasskube UI and either open the ArgoCD UI there, "+
+			"or with the command `glasskube open argo-cd` – but of course you can also use the Argo CLI.\n"+
+			"Follow the ArgoCD docs to get and reset the password to log in: "+
+			"https://argo-cd.readthedocs.io/en/stable/getting_started/#4-login-using-the-cli)")
```
kosmoz, line 142: you can use a raw string constant for this, I think it would look cleaner. see https://github.com/glasskube/glasskube/blob/62f7e830a4bfdfaadfe184534295e96e4700a92f/pkg/bootstrap/bootstrap.go#L65 for example

cmd/glasskube/cmd/output_options.go
```diff
@@ -3,39 +3,46 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/glasskube/glasskube/internal/clientutils"
 	"github.com/spf13/cobra"
 )
 
-type OutputFormat string
+type outputFormatValue string
```
kosmoz, line 10: any reason why it is not called outputFormat?

config/webhook-cert/job.yaml
```diff
@@ -3,6 +3,8 @@ kind: Job
 metadata:
   name: webhook-cert-init
   namespace: system
+  annotations:
+    argocd.argoproj.io/sync-options: Force=true,Replace=true
```
kosmoz, line 7: We could also add this annotation during bootstrap (preprocessManifests). This would have the advantage that we can only add it if "gitops" mode is enabled. What do you think?

internal/clientutils/output.go
```diff
@@ -0,0 +1,113 @@
...
+
+func pruneNonGlasskubeKeys(original map[string]string) map[string]string {
+	res := make(map[string]string)
+	for k, v := range original {
+		if strings.Contains(k, "packages.glasskube.dev") {
+			res[k] = v
+		}
+	}
+	return res
+}
```
kosmoz, line 99: To be honest I still have mixed feelings about not showing the entire object by default, but even removing labels unless explicitly requested seems kind of excessive IMO :confused:

internal/web/response.go
```diff
@@ -46,3 +46,18 @@ func (s *server) swappingRedirect(w http.ResponseWriter, path string, target str
 	locationJson, _ := json.Marshal(locationData)
 	w.Header().Add("Hx-Location", string(locationJson))
 }
+
+func (s *server) sendYamlModal(w http.ResponseWriter, obj string, err error) {
```
kosmoz, line 50: If I understand correctly, `err` here does _not_ represent an error that happened during YAML marshalling, correct? If that is the case, I think this should be made more explicit somehow because initially that's what I assumed this parameter was for.

kosmoz (changes requested):

config/webhook-cert/job.yaml
christophenne: Yes it's a good idea, will do.

internal/clientutils/output.go
christophenne: I agree, but coming from the point of view of a "glasskube argocd starter", isn't it quite a stretch to assume the user needs a super simple do-it-all automation on the one hand, but on the other hand expect them to know which labels and which fields to put into the git repo, and which not? That goes from super easy to reasonably difficult very quick. But maybe that's how it is – I don't know how this could be resolved to be honest. What I would not do is to separate the behaviour between CLI and UI, because they should always behave the same way and offer the same experience.

config/webhook-cert/job.yaml
christophenne: Ok, not so sure anymore. Now if somebody updates with the dry-run bootstrap, it doesn't know whether it is in gitops mode or not? So it would remove the annotation.

cmd/glasskube/cmd/bootstrap_gitops.go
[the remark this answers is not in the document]
pmig: as @kosmoz said Flux is doing it similarly and we should define the parameter as required. It just gives us the option to add more parameter / subcommands in the future :-)

pmig (approved):

config/webhook-cert/job.yaml
kosmoz: We could store this information as a setting via namespace annotation

internal/clientutils/output.go
kosmoz: I understand that, but at the same time it makes any automation that relies on the output less obvious, e.g. `glasskube install foo -o json | jq "..."`.

I guess, all I'm saying is that this change is based on some big assumptions about how the software is used in practice and I kind of challenge the notion that the "gitops" use-case is the most common one. In the end, I think we can go forward with this change but we should keep our ears open for feedback and be prepared to change it again down the line. :shrug:

config/webhook-cert/job.yaml
christophenne: Thanks, yes we already do that but I forgot about it

kosmoz (approved):
Apart from the unresolved conversation above, I think we can merge this. Let's find a resolution there and then :green_circle:  :green_circle:  :green_circle:

kosmoz (approved):

pmig (approved):

internal/clientutils/output.go
pmig: Let's focus on the GitOps case for now

Issue #756, opened 2024-06-05
Possibility to set cache time or to clear caches manually


christophenne:
The default cache time of the repo cache is 5 minutes and is hardcoded. We should discuss strategies for local development and testing where one might want to disable the cache, or reduce the time, or clear it manually with a button click / command.

kosmoz:
Maybe as part of `PackageRepository` spec?

rakshitgondwal:
Hey Folks! Is this open for grabs?

christophenne:
hey @rakshitgondwal sorry, unfortunately it is not at this point.

PR #1054, opened 2024-07-30
fix: add repoclient invalidation

### LINKED ISSUE #988: Client settings on glasskube repo client does not update cache
**Describe the bug**
When adding/updating repo URL using the repo client `glasskube repo add <repoName> <repoURL>` The client settings cache does not update automatically (perhaps only on reconciliation) leading to the `glasskube repo list` command to show a repo URL that is not actually being fetched.

@kosmoz has identified where the logic has to be added to reset repo client cache upon adding new package repositories.  
 
**To Reproduce**

**Expected behavior**
To be able to add, delete and add a new repo URL under the same repo name and to always fetch the latest configured URL. 

**Screenshots**
image (1)

 - Type: [e.g. Kubernetes, OpenShift]
 - Version: [e.g. 1.29]
 - Provider: [e.g. AWS, minikube]

 - Package Operator version: [e.g. v0.0.1]
 - CLI version: [output of `glasskube --version`]

**Additional context**

kosmoz:
Closes #988 

## 📑 Description
This PR adds checking of the repo spec when requesting a repo client from the clientset.
When requesting via name, it checks if the repo spec has been checked in the last 30 seconds (default) and checks again if it is older than that.

## ✅ Checks

### DIFF
--- a/internal/repo/client/clientset.go
+++ b/internal/repo/client/clientset.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"sync"
 	"time"
 
@@ -22,7 +23,8 @@ type defaultClientsetClient struct {
 
 type defaultClientset struct {
 	client            defaultClientsetClient
-	clients           map[string]RepoClient
+	clients           map[string]repoClientWithState
+	clientAge         map[string]time.Time
 	repoWithNameMutex sync.Mutex
 	repoMutex         sync.Mutex
 	maxCacheAge       time.Duration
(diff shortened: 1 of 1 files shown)


internal/repo/client/clientset.go
```diff
@@ -22,24 +23,40 @@ type defaultClientsetClient struct {
...
+	return s.lastCheckedRepoSpec.Add(maxAge).After(time.Now())
+}
+
+func (s *repoClientWithState) checkRepoSpec(repo v1alpha1.PackageRepository) bool {
+	return reflect.DeepEqual(s.repo.Spec, repo.Spec)
+}
+
 func NewClientset(pkgClient adapter.PackageClientAdapter, k8sClient adapter.KubernetesClientAdapter) RepoClientset {
-	return NewClientsetWithMaxCacheAge(pkgClient, k8sClient, 5*time.Minute)
+	return NewClientsetWithMaxCacheAge(pkgClient, k8sClient, 30*time.Second)
```
christophenne, line 50: IMHO 30 seconds is quite short. I know 5 minutes on the other hand is quite long. I think we should do #756.

christophenne (approved):
Works fine :+1:

internal/repo/client/clientset.go
kosmoz: Ah I missed that this value is also used for invalidating the cache for the repo client (not just the repo info). I think it would make sense to allow specifying different values for both. I think 30s is reasonable for the repo info but definitely on the shorter side for the repo client

kosmoz: As for #756, I think it should be tackled in a separate PR as it seems kind of unrelated

kosmoz: updated, let me know what you think

christophenne: > As for https://github.com/glasskube/glasskube/issues/756, I think it should be tackled in a separate PR as it seems kind of unrelated

yes, that's what i meant

christophenne (approved):

PR #1055, opened 2024-07-31
feat: open entrypoint with custom hostname

### LINKED ISSUE #1039: Redirect UI open button to serve alternative IP if --host flag is used
A recently shipped [feature](https://github.com/glasskube/glasskube/pull/1027) allows for the use of the `--host` flag to be appended to `glasskube serve` which serves the Glasskube UI on a host different to the default `localhost`. 

The `Open` button  in the package UI is still associated with localhost (even when the `--host` flag is in use) and therefore is rendered unusable if running glasskube on a remote browserless VM. 

For the `Open` button to be associated to whichever IP that is specified by the `--host` flag. To allow for normal entrypoint access even on remote VMs

christophenne:
Closes #1039

## ✅ Checks

### DIFF
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -42,7 +42,7 @@ func NewOpener() *opener {
 }
 
 func (o *opener) Open(
-	ctx context.Context, pkg ctrlpkg.Package, entrypointName string, port int32) (*OpenResult, error) {
+	ctx context.Context, pkg ctrlpkg.Package, entrypointName string, host string, port int32) (*OpenResult, error) {
 
 	if err := o.initFromContext(ctx); err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


kosmoz (changes requested):
This implementation doesn't support an important convention (which the `--host` flag for `serve` does support): specifying an empty host means listening on all interfaces. here, it looks like it is going to default to "localhost" instead

kosmoz:
Because it doesn't look like the `portforward` package supports this behavior (even though the stdlib `net` package does), one way to emulate it would be using "localhost" as a default value for the host flag and defaulting to "0.0.0.0" in the `open` package.

pkg/open/open.go
```diff
@@ -304,10 +312,24 @@ func getLocalPort(entrypoint v1alpha1.PackageEntrypoint) int32 {
...
+}
+
+func getBrowserUrl(host string, entrypoint v1alpha1.PackageEntrypoint) string {
+	if host == "" {
+		host = "localhost"
+	}
 	url := url.URL{
 		Scheme: "http",
-		Host:   fmt.Sprintf("localhost:%v", getLocalPort(entrypoint)),
+		Host:   fmt.Sprintf("%v:%v", host, getLocalPort(entrypoint)),
```
kosmoz, line 332: could be replaced with `net.JoinHostPort`

kosmoz (approved):

PR #1069, opened 2024-08-05
feat(cli): add --dry-run to uninstall command

### LINKED ISSUE #1061: `glasskube uninstall` should support `--dry-run`
Just as many other commands, `uninstall` should have a `--dry-run` option to simulate package uninstallation. 

**Additional context**
#629

hanshal101:
Closes #1061 
Fixes #1061 

## 📑 Description
This PR adds the `DryRun` options to the uninstall command

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/uninstall.go
+++ b/cmd/glasskube/cmd/uninstall.go
@@ -17,6 +17,7 @@ var uninstallCmdOptions = struct {
 	Yes    bool
 	KindOptions
 	NamespaceOptions
+	DryRunOptions
 }{
 	KindOptions: DefaultKindOptions(),
 }
@@ -67,13 +68,13 @@ var uninstallCmd = &cobra.Command{
 		}
 
 		if uninstallCmdOptions.NoWait {
-			if err := uninstaller.Uninstall(ctx, pkg); err != nil {
+			if err := uninstaller.Uninstall(ctx, pkg, uninstallCmdOptions.DryRun); err != nil {
 				fmt.Fprintf(os.Stderr, "\n❌ An error occurred during uninstallation:\n\n%v\n", err)
 				cliutils.ExitWithError()
 			}
 			fmt.Fprintln(os.Stderr, "Uninstallation started in background")
 		} else {
-			if err := uninstaller.UninstallBlocking(ctx, pkg); err != nil {
+			if err := uninstaller.UninstallBlocking(ctx, pkg, uninstallCmdOptions.DryRun); err != nil {
 				fmt.Fprintf(os.Stderr, "\n❌ An error occurred during uninstallation:\n\n%v\n", err)
 				cliutils.ExitWithError()
 			}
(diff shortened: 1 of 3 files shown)


christophenne (dismissed):
@hanshal101 please test your implementation again, when running `glasskube uninstall cyclops --dry-run`, the command never finishes or seems to wait forever.

hanshal101:
Hey @christophenne sorry for the late update, this error was caused in the `awaitDeletion` function, here I have passed the `isDryRun` bool to check if its true then return nil.

pkg/uninstall/uninstall.go
```diff
@@ -30,28 +30,31 @@ func (obj *uninstaller) WithStatusWriter(sw statuswriter.StatusWriter) *uninstal
...
 	obj.status.Start()
 	defer obj.status.Stop()
-	err := obj.delete(ctx, pkg)
+	err := obj.delete(ctx, pkg, isDryRun)
 	if err != nil {
 		return err
 	}
-	return obj.awaitDeletion(ctx, pkg)
+
+	return obj.awaitDeletion(ctx, pkg, isDryRun)
```
kosmoz, line 41:
```suggestion
	if isDryRun {
		return nil
	} else {
		return obj.awaitDeletion(ctx, pkg)
	}
```

pkg/uninstall/uninstall.go
```diff
@@ -64,7 +67,11 @@ func (uninstaller *uninstaller) delete(ctx context.Context, pkg ctrlpkg.Package)
 	}
 }
 
-func (obj *uninstaller) awaitDeletion(ctx context.Context, pkg ctrlpkg.Package) error {
+func (obj *uninstaller) awaitDeletion(ctx context.Context, pkg ctrlpkg.Package, isDryrun bool) error {
+	if isDryrun {
+		return nil
+	}
+
```
kosmoz, line 74:
```suggestion
func (obj *uninstaller) awaitDeletion(ctx context.Context, pkg ctrlpkg.Package) error {
```

kosmoz (changes requested):
I think it would be better to keep the logic out of the await function (see suggestions below).

Also, I think there should be some feedback that dry run mode is enabled. See for example the configure command: https://github.com/hanshal101/glasskube/blob/26abda901fab05ceee737a99c710997f48b6c453/cmd/glasskube/cmd/configure.go#L51-L53

kosmoz:
> Also, I think there should be some feedback that dry run mode is enabled. See for example the configure command: https://github.com/hanshal101/glasskube/blob/26abda901fab05ceee737a99c710997f48b6c453/cmd/glasskube/cmd/configure.go#L51-L53

Please add some feedback that dry-run is enabled.

hanshal101:
Code updated !!!

cmd/glasskube/cmd/uninstall.go
```diff
@@ -66,14 +67,18 @@ var uninstallCmd = &cobra.Command{
 			}
 		}
 
+		if uninstallCmdOptions.DryRun {
+			fmt.Fprintln(os.Stderr, "🔎 Dry-run mode is enabled. Nothing will be changed.")
+		}
+
```
kosmoz, line 73: Please move this so that it is shown before the uninstall details and confirmation prompt

kosmoz (approved):

PR #1089, opened 2024-08-08
docs: gitops integration

christophenne:
## ✅ Checks


pmig (changes requested):
We should also document the git subcommand in the bootstrap command documentation, else it looks good. On point :-)

pmig (approved):

PR #1094, opened 2024-08-10
feat(package-operator): add package repository validating webhook

### LINKED ISSUE #871: Webhook Validator for `PackageRepository` deletion
`PackageRepository` resources can be deleted, even if there are packages installed from that repository.

There should be a webhook validator in the operator, that, when a `PackageRepository` is about to be deleted, should check whether there are no packages installed from that repository. There should be a validation error, if packages from the repo are still installed. 

**Additional context**
There are already some webhook validations in place, e.g. for `ClusterPackage` resources.

botchk:
Closes #871 

## 📑 Description
Adds a validating webhook for the package repository resource
* Upon deletion of a repository, or update of the URL of a repository, the following validation is performed:
  * If no packages are installed, the delete/update proceeds as normal. If packages are installed, an error with the list of installed packages is shown
  * ```
    k delete packagerepositories.packages.glasskube.dev test
    Error from server (Forbidden): admission webhook "vpackagerepository.kb.io" denied the request: dependent package(s) installed: [{ingress-nginx v1.9.5+1 test}]
    ```

## ✅ Checks

## ℹ Additional Information
<!-- Any additional information like breaking changes, dependencies added, screenshots, comparisons between new and old behavior, etc. -->

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -37,6 +37,9 @@ resources:
   kind: PackageRepository
   path: github.com/glasskube/glasskube/api/v1alpha1
   version: v1alpha1
+  webhooks:
+    validation: true
+    webhookVersion: v1
 - api:
     crdVersion: v1
     namespaced: true
--- a/cmd/cert-manager/main.go
+++ b/cmd/cert-manager/main.go
@@ -21,7 +21,7 @@ var (
 	webhookConfigName = "glasskube-validating-webhook-configuration"
 	namespace         = "glasskube-system"
 	certDir           = ""
-	webhookNames      = []string{"vpackage.kb.io", "vclusterpackage.kb.io"}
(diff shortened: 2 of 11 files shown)


yeahservice:
@hanshal101 I've added the validating webhook skeleton. Next up would be the implementation of checking for installed packages from this repository (+tests). Let me know if you want to do this.

We should also consider the use case when the URL  from an repository is updated. In this case the packages offered by the repository can completely change. So we probably also shouldn't allow this when packages from this repository are currently installed.

yeahservice:
@christophenne I've added a step by step guide on how to add a new webhook, since there are quite some non-trivial steps involved. Could we add this as guide in the contributing section somewhere? Or at least keep it as comment in the PR to link here for future webhook implementations.

kosmoz (approved):
Good work, LGTM!

Issue #1102, opened 2024-08-13
Add `--delete-namespace` flag to uninstall command


kosmoz:
**Is your feature request related to a problem? Please describe.**
It can be frustrating to have lots of dangling namespaces in a cluster that are no longer in use. Glasskube should offer an option to delete the namespace when uninstalling a package.

**Describe the solution you'd like**
The `uninstall` command can be invoked with a `--delete-namespace` flag. If this is the case, it should check if there are any Packages installed in that namespace, other than the one that should be uninstalled. If there are, it should print an error message and abort the operation, otherwise it should proceed to delete the `Package` and subsequently the namespace resource.

**Additional context**
For context, see #1084.
Related to: #1101

hanshal101:
Hey @kosmoz I can work on this !!!

Pavan-Chaithanya:
Hi @kosmoz, I would like to work on this

kosmoz:
Glad to hear that! Let us know if you need anything.

anishbista60:
@Pavan-Chaithanya Are you working in this isssue ?

Pavan-Chaithanya:
> @Pavan-Chaithanya Are you working in this isssue ?

Yes, I am working on it

Baalekshan:
@christophenne !

hanshal101:
Hey @christophenne is this issue still relevent?

Baalekshan:
Hey @hanshal101 , the PR is complete I am waiting for approval.

hanshal101:
> Hey @hanshal101 , the PR is complete I am waiting for approval.

Ohh cool!

PR #1111, opened 2024-08-15
feat(cli): add --delete-namespace to delete namespace incase it is empty after package deletion

Pavan-Chaithanya:
Closes https://github.com/glasskube/glasskube/issues/1102

## 📑 Description
This PR introduces a new flag `--delete-namespace` to the `uninstall` command. When this flag is passed to uninstall a package, the following can happen:
1. If no package is present in the namespace other than the package we are trying to uninstall, It uninstalls the package and deletes the namespace.
2. If other packages are present in the namespace, it returns an error to the CLI. In this case, you might wanna consider deleting the package without this flag.

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/uninstall.go
+++ b/cmd/glasskube/cmd/uninstall.go
@@ -7,14 +7,16 @@ import (
 	"github.com/fatih/color"
 	"github.com/glasskube/glasskube/internal/clicontext"
 	"github.com/glasskube/glasskube/internal/cliutils"
+	"github.com/glasskube/glasskube/pkg/list"
 	"github.com/glasskube/glasskube/pkg/statuswriter"
 	"github.com/glasskube/glasskube/pkg/uninstall"
 	"github.com/spf13/cobra"
 )
 
 var uninstallCmdOptions = struct {
-	NoWait bool
-	Yes    bool
+	NoWait          bool
+	Yes             bool
+	DeleteNamespace bool
 	KindOptions
 	NamespaceOptions
 }{
@@ -46,6 +48,7 @@ var uninstallCmd = &cobra.Command{
 			cliutils.ExitWithError()
 		}
 
+		var deleteNamespace bool
 		if !pkg.IsNamespaceScoped() {
 			if g, err := dm.NewGraph(ctx); err != nil {
 				fmt.Fprintf(os.Stderr, "❌ Error validating uninstall: %v\n", err)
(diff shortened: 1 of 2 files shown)


cmd/glasskube/cmd/uninstall.go
```diff
@@ -64,16 +67,39 @@ var uninstallCmd = &cobra.Command{
...
+			pkgCount := 0
+			for _, p := range pkgsInNamespace {
+				pkgCount += len(p.Packages)
+			}
+			if pkgCount > 1 {
+				fmt.Fprintf(os.Stderr, "❌ Namespace %v contains more than one package. "+
+					"Current count of packages present in the namespace: %v\n", namespace, pkgCount)
+				cliutils.ExitWithError()
+			}
+			deleteNamespace = true
```
kosmoz, line 90: This seems like a lot of unnecessary work. It would be better to use a simple packageClient and just fetch the Package resources directly, rather than using the list package, since that one goes through some extra hoops.

kosmoz (changes requested):

pkg/uninstall/uninstall.go
```diff
@@ -31,23 +32,43 @@ func (obj *uninstaller) WithStatusWriter(sw statuswriter.StatusWriter) *uninstal
...
-	defer obj.status.Stop()
-	return obj.delete(ctx, pkg)
+// Uninstall deletes the v1alpha1.Package custom resource from the cluster
+// and namespace (if empty).
+func (obj *uninstaller) UninstallAndDeleteNamespace(
+	ctx context.Context,
+	pkg ctrlpkg.Package,
+	deleteNamespace bool,
+) error {
+	return obj.Uninstall(ctx, pkg)
```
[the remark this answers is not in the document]
Pavan-Chaithanya: Removed it

cmd/glasskube/cmd/uninstall.go
Pavan-Chaithanya: Makes sense. Fixed it

Pavan-Chaithanya:
@kosmoz Can you please review this PR?

pkg/uninstall/uninstall.go
```diff
@@ -52,6 +54,32 @@ func (obj *uninstaller) Uninstall(ctx context.Context, pkg ctrlpkg.Package, isDr
...
+	isDryRun bool,
+) error {
+	namespace := pkg.GetNamespace()
+	if err := obj.UninstallBlocking(ctx, pkg, isDryRun); err != nil {
+		return err
+	}
+	clientset := clicontext.KubernetesClientFromContext(ctx)
+
+	return obj.deleteNamespaceBlocking(ctx, clientset, namespace)
+}
```
kosmoz, line 81: I think it would be simpler to add a "DeleteNamespace" option to the existing UninstallBlocking function

pkg/uninstall/uninstall.go
```diff
@@ -52,6 +54,32 @@ func (obj *uninstaller) Uninstall(ctx context.Context, pkg ctrlpkg.Package, isDr
...
+	pkg ctrlpkg.Package,
+	isDryRun bool,
+) error {
+	namespace := pkg.GetNamespace()
+	if err := obj.UninstallBlocking(ctx, pkg, isDryRun); err != nil {
+		return err
+	}
+	clientset := clicontext.KubernetesClientFromContext(ctx)
+
+	return obj.deleteNamespaceBlocking(ctx, clientset, namespace)
```
kosmoz, line 80: clientset should be initialized in deleteNamespaceBlocking

cmd/glasskube/cmd/uninstall.go
```diff
@@ -78,11 +79,26 @@ var uninstallCmd = &cobra.Command{
 			}
 			fmt.Fprintln(os.Stderr, "Uninstallation started in background")
 		} else {
-			if err := uninstaller.UninstallBlocking(ctx, pkg, uninstallCmdOptions.DryRun); err != nil {
-				fmt.Fprintf(os.Stderr, "\n❌ An error occurred during uninstallation:\n\n%v\n", err)
-				cliutils.ExitWithError()
+			// delete the namespace only if it is namespace scoped and there are no other packages in the namespace since
+			// cluster packages delete the namespace once they are deleted
+			deleteNamespace := uninstallCmdOptions.DeleteNamespace && pkg.IsNamespaceScoped()
```
kosmoz, line 84: If DeleteNamespace is set and the package is not namespace-scoped, the command shoul exit with an error

cmd/glasskube/cmd/uninstall.go
```diff
@@ -78,11 +79,26 @@ var uninstallCmd = &cobra.Command{
...
+			if deleteNamespace {
+				if err := uninstaller.IsNamespaceSafeToDelete(ctx, pkg); err != nil {
```
kosmoz, line 86: I think it might be possible to perform this check in the UninstallBlocking method

kosmoz (changes requested):

kosmoz:
Hey @Pavan-Chaithanya, how are you doing? Do you need any help with this PR?

Pavan-Chaithanya:
Hey @kosmoz, I'm extremely sorry about the delay. I got occupied with other things. I addressed all the comments. Can you have a look? Btw I made some changes in makefile, Reason:
Previously, we ran the lint with whatever version was installed in bin/. But if one of your team upgrades the version of lint, the user cannot automatically upgrade unless he sees it and manually deletes bin/ and installs it. With this change, we check the version and see if a new version needs to be installed and install it.

Pavan-Chaithanya:
> Hey @kosmoz, I'm extremely sorry about the delay. I got occupied with other things. I addressed all the comments. Can you have a look? Btw I made some changes in makefile, Reason: Previously, we ran the lint with whatever version was installed in bin/. But if one of your team upgrades the version of lint, the user cannot automatically upgrade unless he sees it and manually deletes bin/ and installs it. With this change, we check the version and see if a new version needs to be installed and install it.

I fixed it because I faced the issue of having an older version and seeing errors which I didn't see on GitHub actions 😓 . You might wanna consider doing the same for the other version if you think this is ok.

Pavan-Chaithanya:
@kosmoz Any updates on this PR?

PR #1129, opened 2024-08-22
feat: add support for package components

### LINKED ISSUE #1126: Add support for package subcomponents
Some packages contain components that are identical or very similar to other packages. There should be a way to reuse these component definitions for multiple packages. A prime example for this would be databases, but other components should be possible as well. 

A package manifest file can include a list of components that should be installed alongside the package. A component declaration can be very similar to a dependency declaration but there are some differences in how they are handled during installation:
1. In contrast to dependencies, components are not shared between different package installations. 
2. Components are always prefixed with the name of the package they are installed as part of.
3. While dependencies are required to be cluster-scoped, components are required to be namespace-scoped. (This requirement follows from the points above and the fact that cluster-scoped packages may include components)

WE also considered adding support for packages that include more than one helm release and then creating helm charts for the components, but this has several disadvantages: 
1. Only helm charts are supported (we'd prefer to avoid to depend on helm more than necessary)
2. The version of a component can not be changed independently of the parent package

**Additional context**
This is sort of a follow-up of #112 but it became more pressing in the light of current package integration efforts.

kosmoz:
Closes #1126 

## 📑 Description
This PR adds support for Packages and ClusterPackages that contain other Packages in the form of "components". 

## ✅ Checks


kosmoz:
TODO: components are not cleaned up on uninstall

PR #1131, opened 2024-08-23
feat(cli): create namespace if necessary when installing a package

### LINKED ISSUE #1101: CLI: Create namespace if necessary when installing a namespace-scoped Package
Installing a namespace-scoped package can be more complicated than necessary if the target namespace does not exist yet. In that case, a user has to resort to `kubectl` or other means to create the namespace before being able to install the package. This should not be necessary.

When installing a namespace-scoped package, the CLI should check if the target namespace exists and, if it does not, create it after the user has confirmed the install operation (if no `--yes` flag is present) and before the `Package` is created. The installation summary should inform the user that the namespace is going to be created. Example output:

```
$ glasskube install quickwit -n test
Summary:
 * The following packages will be installed in your cluster (minikube):
    1. quickwit of type quickwit in namespace test (version v0.8.1+5)
 * Namespace test does not exist and will be created
 * Automatic updates will be not enabled
```

**Additional context**
For context, see #1084

hanshal101:
Closes #1101 
Fixes #1101

## 📑 Description
This PR fixes the bug Creating namespace if necessary when installing a namespace-scoped Package.

## ✅ Checks

## ℹ Additional Information
My `list` got broke after solving this, it looks something like this, need help here:
image


kosmoz:
> My `list` got broke after solving this, it looks something like this, need help here:

What do you mean when you say it's broken? It looks fine to me

hanshal101:
Ohh so I think its just with me I guess, I'm sorry my terminal size was small hence was not showing all details

hanshal101:
Hey, just making an update in the PR. I'll take a week to solve this !!! sorry for the inconvenience

hanshal101:
Just made all the changes here, while I was not clear with adding the `namespace check and install` in the install function.

hanshal101:
> need this sooner and I took this over and fixed and the

Cool, Sorry for the delay though

cmd/glasskube/cmd/install.go
```diff
@@ -219,6 +223,18 @@ var installCmd = &cobra.Command{
 			fmt.Fprintln(os.Stderr, " * Automatic updates will be", bold("not enabled"))
 		}
 
+		createNamespace := false
+		if installCmdOptions.NamespaceOptions.Namespace != "" {
+			if ok, err := namespaces.IsNamespaceInstalled(ctx, cs, installCmdOptions.NamespaceOptions.Namespace); !ok {
+				fmt.Printf(" * Namespace %v does not exist and will be created\n",
```
kosmoz, line 229:
```suggestion
				fmt.Fprintf(os.Stderr, " * Namespace %v does not exist and will be created\n",
```

internal/namespaces/ultils.go
```diff
@@ -0,0 +1,38 @@
...
+		},
+	}
+
+	_, err := cs.CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
kosmoz, line 38: I'm not convinced this function adds much on top of just calling `Create` directly

internal/namespaces/ultils.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func IsNamespaceInstalled(ctx context.Context, cs *kubernetes.Clientset, namespace string) (bool, error) {
```
kosmoz, line 12: Personally, I would change the name to simply `Exists`, since
1. A namespace is not installed and
2. the package is already called namespaces

kosmoz (changes requested):

kosmoz (approved):

Issue #1124, opened 2024-08-19
GlassKube should handle errors gracefully while a package is being installed.


vbkotecha:
**Describe the bug**

**To Reproduce**
1. Click "Install ArgoCD" through the UI
2. Click Open while the installation is not yet complete (this is often the case for impatient DevOps engineers.)
3. You will see an error in the UI and the CLI that says `failed to open argo-cd: could not open entrypoint [anonymous]: no pod found for service argocd-server has status ready`

**Expected behavior**

**Screenshots**
<img width="400" alt="Screenshot 2024-08-19 at 5 22 29 PM" src="https://github.com/user-attachments/assets/cc6f77c1-e505-4212-88bd-6d6d5d8f737b">

**Cluster Info (please complete the following information):**
 - Type: Kubernetes
 - Version: 1.30
 - Provider: Docker Desktop

**Glasskube Info (please complete the following information):**
 - Package Operator version:  0.17.0
 - CLI version: 0.17.0

kosmoz:
We have considered adding a "status probes" feature to packages for a while and I think this is one of the issues that could be resolved very easily with those. By "status probe" I mean adding a set of (resource ID; JSON path; expected value) tuples to a package manifest. The operator could then evaluate those probes by fetching the resource and checking the value at the given path against the expected value. Then we could easily only show the open button if all status probes of a package are OK.

Do you have any thoughts about this @vbkotecha?

PR #1154, opened 2024-09-02
feat(cli): improve gitops bootstrap support

### LINKED ISSUE #1173: No matches for PackageRepository error should not appear in gitops bootstrap
The PackageRepository custom resource definition can not be found, because it doesn't exist yet. But this error is catchable, because we can assume that when the CRD does not exist, there also simply exists no resource of that type. So we don't have to print that error to the console and might be able to do the bootstrap without the force argument (?).

christophenne:
Closes #1173 
Ref https://github.com/glasskube/gitops-template/issues/15

## 📑 Description
* Added support for private GitHub repositories. Repo credentials are passed via `--username` and `--token` arguments. 
* Made the command block until argo-cd is actually openable, and print the initial admin password. 

## ✅ Checks

## ℹ Additional Information
https://github.com/glasskube/gitops-template/pull/17

### DIFF
--- a/cmd/glasskube/cmd/bootstrap_git.go
+++ b/cmd/glasskube/cmd/bootstrap_git.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -155,10 +205,49 @@ var bootstrapGitCmd = &cobra.Command{
...
+		for {
+			select {
+			case <-ctx.Done():
+				fmt.Fprintf(os.Stderr, "\nargo-cd has not become ready in the specified timeout.\n")
+				break WaitLoop
+			case <-tickC:
+				if res, err := open.NewSilentOpener().Open(ctx, argoCdPkg, "", "localhost", 0); err == nil {
+					res.Stop()
+					break WaitLoop
+				}
```
kosmoz, line 226: I think a more lightweight approach could be extracting the code that determines the active pod in the `open` package to a separate function and using that instead. 

This could also be used as an easy solution for https://github.com/glasskube/glasskube/issues/1124. Once we implement the probes outlined in the comment below the above issue we can refactor the gitops bootstrap to use those as well

cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -69,8 +84,18 @@ var bootstrapGitCmd = &cobra.Command{
 
 		// regularly bootstrap in the cluster
 		// without a git client, this means it only supports GitHub for now
-		basePath := strings.ReplaceAll(bootstrapGitCmdOptions.url, "github.com", "raw.githubusercontent.com")
-		basePath = basePath + "/main"
+		host := "raw.githubusercontent.com"
+		if bootstrapGitCmdOptions.token != "" {
+			// might also be GitHub specific? a more proper way for the future could be to send it via the
+			// Authorization header instead, but that would also have to be supported by the BootstrapClient
+			host = bootstrapGitCmdOptions.token + "@" + host
```
kosmoz, line 91: I am somehow amazed that this even works, lol

But is the username even needed, apart from saving it to the secret below?

cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -155,10 +205,49 @@ var bootstrapGitCmd = &cobra.Command{
...
 
 func init() {
-	bootstrapGitCmd.Flags().StringVar(&bootstrapGitCmdOptions.url, "url", "", "URL of the GitOps Repository")
+	bootstrapGitCmd.Flags().StringVar(&bootstrapGitCmdOptions.url, "url", "",
+		"URL of the GitOps Repository")
+	bootstrapGitCmd.Flags().StringVar(&bootstrapGitCmdOptions.branch, "branch", "",
+		"Branch of the GitOps Repository to use (default is main)")
+	bootstrapGitCmd.Flags().StringVar(&bootstrapGitCmdOptions.username, "username", "",
+		"Username to use for authentication")
+	bootstrapGitCmd.Flags().StringVar(&bootstrapGitCmdOptions.token, "token", "",
```
kosmoz, line 251: A lot of tools also support the `GITHUB_TOKEN` environment variable. I think we should do the same.

cmd/glasskube/cmd/bootstrap_git.go
```diff
@@ -155,10 +205,48 @@ var bootstrapGitCmd = &cobra.Command{
...
+			}
+		}
+
+		// get initial admin password to print it
+		var adminPwStr string
+		if adminPw, err := clientset.CoreV1().Secrets(argocdNamespace).
+			Get(ctx, "argocd-initial-admin-secret", metav1.GetOptions{}); err != nil {
+			fmt.Fprintf(os.Stderr, "\nCould not obtain initial ArgoCD password: %v\n\n", err)
+		} else {
+			adminPwStr = string(adminPw.Data["password"])
```
kosmoz, line 236: I think secret data must be deserialized from base64, for example see https://github.com/glasskube/glasskube/blob/71f12d3ba10239d0197aab1a2315f6c970269d69/internal/manifestvalues/resolver.go#L80

cmd/glasskube/cmd/bootstrap_git.go
kosmoz: I thought the username can be removed entirely, but on closer inspection it appears that it is needed for argo-cd to access the repo (?) 

If it can not be removed feel free to mark this conversation as resolved

kosmoz (commented):
It works great! Two things I noticed:
* As discussed, the token is printed in the logs, which is not nice. you can parse the URL with `url.Parse()` and then use `parsedUrl.Redacted()` for printing to prevent this, but you have to change the url to  `$USER:$TOKEN@...` for that to work
* It would be cool if URLs ending with ".git" would be supported as well, since that's the default when you use the GitHub UI to copy the repo URL. But I don't think it's necessary to do this in this PR, since it's not a regression

kosmoz (approved):

PR #1166, opened 2024-09-03
Use default namespace when --yes is specified 

### LINKED ISSUE #1163: `glasskube install` should use default name if `--yes` is specified
When a user specifies the `--yes` flag for a CLI command, they generally want to avoid interactivity. Glasskube, however, always asks the user to enter a name for the package. 

When `--yes` is specified, the default name (same as the package name) should be used without asking.

**Additional context**
This will help us greatly for our package CI pipelines, see e.g. https://github.com/glasskube/packages/pull/380

NIXBLACK11:
Closes #1163

## 📑 Description
Using the default namespace when the --yes flag is specified with the `glasskube install` command.
If `installCmdOptions.Yes` is set the default namespace is used.

## ✅ Checks

### DIFF
--- a/cmd/glasskube/cmd/install.go
+++ b/cmd/glasskube/cmd/install.go
@@ -147,10 +147,14 @@ var installCmd = &cobra.Command{
 		} else {
 			var name string
 			if len(args) != 2 {
-				fmt.Fprintf(os.Stderr, "%v has scope Namespaced. Please enter a name (default %v):\n", packageName, packageName)
-				name = cliutils.GetInputStr("name")
-				if name == "" {
+				if installCmdOptions.Yes {
 					name = packageName
+				} else {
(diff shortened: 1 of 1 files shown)


cmd/glasskube/cmd/install.go
```diff
@@ -147,10 +147,14 @@ var installCmd = &cobra.Command{
 		} else {
 			var name string
 			if len(args) != 2 {
-				fmt.Fprintf(os.Stderr, "%v has scope Namespaced. Please enter a name (default %v):\n", packageName, packageName)
-				name = cliutils.GetInputStr("name")
-				if name == "" {
+				if installCmdOptions.Yes {
```
kosmoz, line 150:
```suggestion
				if installCmdOptions.Yes {
					fmt.Fprintf(os.Stderr, "Name not specified. Using default name: %v\n", packageName)
```

kosmoz (changes requested):

kosmoz (approved):

PR #1172, opened 2024-09-04
fix: retry on certificate errors during bootstrap

### LINKED ISSUE #1162: Retry applying default package repo during bootstrap in existing installations
As #1155 has shown, during bootstrap the cert-manager job is recreated and new certificates are issued. This can lead to a short time span in the bootstrap process, where certificate errors occur. 

There should be a retry logic when applying the default repository, in case the error is a certificate error like shown in the issue. 

**Additional context**
#1155 #1160

christophenne:
Closes #1162

## ✅ Checks

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
+
 	"github.com/fatih/color"
 	"github.com/glasskube/glasskube/api/v1alpha1"
 	"github.com/glasskube/glasskube/internal/clientutils"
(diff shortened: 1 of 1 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -258,20 +260,40 @@ func (c *BootstrapClient) applyManifests(
...
+			// failed calling webhook "vpackagerepository.kb.io": failed to call webhook: Post "https://.../validate-...":
+			// tls: failed to verify certificate: x509: certificate signed by unknown authority
+			// however, there can apparently also be other errors like "resource not found" wrapped in the internal error,
+			// so instead of explicitly looking for the cert error message, we simply retry in all internal error cases
+			if err = retry.OnError(wait.Backoff{
+				Duration: 2 * time.Second,
+				Factor:   3,
+				Jitter:   0.1,
+				Steps:    5,
+			}, apierrors.IsInternalError, applyFn); err != nil {
```
kosmoz, line 281: Is there a reason for not using `retry.DefaultRetry`? Because if that would also be possible, we could simplify this to (something like):
```go
retry.OnError(
  retry.DefaultRetry, 
  func(err error) bool { return apierrors.IsInternalError(err) || apierrors.IsConflict(err) },
  applyFn,
)
```

… and then possibly inline applyFn

kosmoz: We could also create a `predicate` package with the following generic function to make this kind of thing more ergonomic:
```
func All[T any](fns ...func(T) bool) func(T) bool {
	return func(arg T) bool {
		for _, fn := range fns {
			if fn(arg) {
				return true
			}
		}
		return false
	}
}
```

kosmoz: OK, but I'm still not a fan of having different retry logic based on the resource kind. E.g. what if there would be a conflict for the `PackageRepository`. I think we could also nest the retry calls like this:

```go
err = retry.OnError(InternalErrorBackoff, apierrors.IsInternalError, 
  retry.RetryOnConflict(retry.DefaultRetry, func() error {
    // ...
  })
)
```

What do you think?

kosmoz (approved):

PR #1181, opened 2024-09-04
feat(ui): add editing of package repositories

### LINKED ISSUE #860: Repository configuration via UI frontend
Currently it is only possible to add and update repository via the `cli`.

When clicking on a repository a detail page should be shown (similar to `ClusterPackages`) where a user can configure the name, url and checkbox if the repo is the default repo.

Additionally there should be button to add a new repository in the top right corner.

The scope of this ticket is to create the html forms in order to send the form request to the backend.

**Additional context**
Related backend issue: https://github.com/glasskube/glasskube/issues/861
Design specification: https://glasskube.dev/docs/design/repositories/

### LINKED ISSUE #861: Repository configuration via UI backend
Currently it is only possible to add and update repository via the `cli`.

When clicking on a repository a detail page should be shown (similar to `ClusterPackages`) where a user can configure the name, url and checkbox if the repo is the default repo.

Additionally there should be button to add a new repository in the top right corner.

The scope of this ticket is to create the golang backend to receive the form request from the frontend and process it similar to how the cli processes it.

**Additional context**
Related frontend issue: https://github.com/glasskube/glasskube/issues/860
Design specification: https://glasskube.dev/docs/design/repositories/

prikshitsingh24:
Closes #861 , #860

## 📑 Description
Adding a separate page linked to setting which allows to edit the repositories. Can edit url and default status of the repo config file.
## ✅ Checks

## ℹ Additional Information
Editing the Url for the repo:
https://github.com/user-attachments/assets/402569bf-8f5b-47ec-942e-b5ac656258f5

Showing error if not able to edit:

https://github.com/user-attachments/assets/ce07dc24-26c7-47aa-a079-0639a0d610ab

### DIFF
--- a/internal/web/templates/pages/repository.html
+++ b/internal/web/templates/pages/repository.html
@@ -1,29 +1,39 @@
 {{ define "content" }}
-<div class="container mx-auto p-4 shadow-md rounded-lg mt-8">
+  <div class="container mx-auto p-4 shadow-md rounded-lg mt-8">
     <h1 class="text-2xl font-bold mb-4">Repository Configuration</h1>
     <form class="space-y-4">
-        <div>
-            <label for="name" class="form-label">Name</label>
-            <div class="input-group mb-2">
-                <input type="text" id="name" name="name" value="{{ .Repository.Name }}" disabled class="form-control" />
-            </div>
+      <div>
+        <label for="name" class="form-label">Name</label>
+        <div class="input-group mb-2">
+          <input type="text" id="name" name="name" value="{{ .Repository.Name }}" disabled class="form-control" />
         </div>
-        <div>
-            <label for="url" class="form-label">URL</label>
-            <div class="input-group mb-2">
-                <input type="text" id="url" name="url" value="{{ .Repository.Spec.Url }}" required class="form-control" />
-            </div>
+      </div>
+      <div>
+        <label for="url" class="form-label">URL</label>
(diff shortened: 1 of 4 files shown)


kosmoz (commented):
This looks really good! I have two suggestions:

1. There is no way to create or delete repositories. I think that's OK for now, we can do it in a later stage, but the text on the settings page needs to be updated: `Please use the CLI to create or delete package repositories: <code>glasskube help repo</code>.`
2. I think it would make sense to validate the URL before submitting. You can try to parse it with `url.Parse` in the endpoint and handle the error appropriately, that should be enough.

internal/web/templates/pages/repository.html
```diff
@@ -0,0 +1,39 @@
...
+        </div>
+      </div>
+      <div>
+        <label for="default" class="form-check mt-1 mb-3">
+          <input
+            type="checkbox"
+            id="default"
+            name="default"
+            {{ if .Repository.IsDefaultRepository }}checked{{ end }}
+            class="form-check-input" />
```
kosmoz, line 24: Unchecking this for the default repo does nothing, which I think is OK. There is not supposed to be no default repo. But I think it would make sense to also disable the checkbox for the current default repo, so that it can not be unchecked.

prikshitsingh24: done

kosmoz (approved):

Issue #1191, opened 2024-09-05
generate completions on homebrew install


totallyGreg:
**Is your feature request related to a problem? Please describe.**
Goreleaser.yaml can define the homebrew builtin functions to generate and install the shell completions in the appropriate directories

**Describe the solution you'd like**
it would be convenient to have shell completion automatically on install with homebrew

**Describe alternatives you've considered**
Currently there are instructions on how to generate and source the completions post install. 

**Additional context**
https://docs.brew.sh/Formula-Cookbook
https://carlosbecker.com/posts/golang-completions-cobra/

```
    install: |
      bin.install "glasskube"
      bash_completion.install "completions/glasskube.bash" => "glasskube"
      zsh_completion.install "completions/glasskube.zsh" => "_glasskube"
      fish_completion.install "completions/glasskube.fish"
```

PR #1192, opened 2024-09-05
build(cli): add completions on brew install

totallyGreg:
Closes # [1191](https://github.com/glasskube/glasskube/issues/1191)

## 📑 Description
add the three homebrew shell completion functions to the goreleaser.yaml brew install block

## ✅ Checks

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -170,6 +170,9 @@ brews:
     commit_msg_template: "chore: brew formula update for {{ .ProjectName }} version {{ .Tag }}"
     install: |
       bin.install "glasskube"
+      bash_completion.install "completions/glasskube.bash" => "glasskube"
+      zsh_completion.install "completions/glasskube.zsh" => "_glasskube"
+      fish_completion.install "completions/glasskube.fish"
     skip_upload: "auto"
 
 release:


pmig (approved):
Thanks enhancing the developer experience

kosmoz:
Should we do the same in the official brew repo?

pmig:
I tried it out and it actually didn't work, so I am reverting the change now

```shell
Error: An exception occurred within a child process:
  Errno::ENOENT: No such file or directory - completions/glasskube.bash
```

It seems that the completions are just copied, we would need to execute the cli completions command and put it into the correct folder. Else we would just create a bunch of files in the home directory for everone.

totallyGreg:
I made a mistake.  There is a homebrew function [generate_completions_from_executable-instance_method](https://rubydoc.brew.sh/Formula.html#generate_completions_from_executable-instance_method) that can be called after install to generate and install the completions.  I have removed the install lines and created a post_install block in the goreleaser.yaml.  I'll send a new pr.

pmig:
As the current installation is broken, I also created https://github.com/glasskube/glasskube/pull/1193 as a fallback.

PR #1194, opened 2024-09-06
feat: generate completions after brew install

totallyGreg:
Resolves issue with Completions introduced in https://github.com/glasskube/glasskube/pull/1192

## 📑 Description
Removes three install lines from previous commit and uses homebrew function in post_install block to generate and install the completions
[Class: Formula      — Homebrew Ruby API](https://rubydoc.brew.sh/Formula.html#generate_completions_from_executable-instance_method)

## ✅ Checks

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -170,6 +170,8 @@ brews:
     commit_msg_template: "chore: brew formula update for {{ .ProjectName }} version {{ .Tag }}"
     install: |
       bin.install "glasskube"
+    post_install: |
+      generate_completions_from_executable(bin/"glasskube", "completions")
     skip_upload: "auto"
 
 release:


pmig (commented):
@totallyGreg can you reference any application in a homebrew tap that has following post_installation script implemented so we can try it out?

I will leave the PR open until then, so we can assure it works.

totallyGreg:
Sidero added the function here:  https://github.com/siderolabs/homebrew-tap/pull/54
And then moved it to the post_install block here:  https://github.com/siderolabs/homebrew-tap/pull/56

pmig:
I installed talosctl and the completions did not work for me (fish shell on linux) do they completions work for you?

totallyGreg:
They do for me in zsh.  
```shell
 [13:06] ❯ brew list talosctl
/opt/homebrew/Cellar/talosctl/1.7.5/bin/talosctl
/opt/homebrew/Cellar/talosctl/1.7.5/etc/bash_completion.d/talosctl
/opt/homebrew/Cellar/talosctl/1.7.5/sbom.spdx.json
/opt/homebrew/Cellar/talosctl/1.7.5/share/fish/vendor_completions.d/talosctl.fish
/opt/homebrew/Cellar/talosctl/1.7.5/share/zsh/site-functions/_talosctl

 [13:09] ❯ whence -v _talosctl
_talosctl is an autoload shell function from /opt/homebrew/share/zsh/site-functions/_talosctl
```

pmig (approved):

pmig:
after I fixed a small typo in https://github.com/glasskube/glasskube/pull/1200 completions now should work. Do they work on your side?

PR #1237, opened 2024-09-13
fix: increase resilience in case of broken repositories

### LINKED ISSUE #976: Can not open package details if there is a package repository that does not work
**Describe the bug**
If there is at least one `PackageRepository` in the cluster that does not work (e.g. server unavailable), it is not possible to show the details page for any package that has not been installed (both cluster-scoped and namespace-scoped).

**To Reproduce**
1. Add a broken repo: `glasskube repo add broken https://not-a-glasskube-repo.org/`
2. Start UI (there should be an error at the top saying that not all repos could be fetched, this is ok)
3. Try to open any package that is not installed

**Expected behavior**
Possible to show package details and install from the working repository.

**Screenshots**
image

 - Type: any
 - Version: any
 - Provider: any

 - Package Operator version: dev (latest main)
 - CLI version: dev (latest main)

**Additional context**

christophenne:
Closes #976

## 📑 Description
While fixing #976, I also came across some other use cases that wouldn't work as soon as at least one repository is not available:
* `glasskube describe`
* `glasskube install` and installation via UI (because of dependency manager)
* package reconciliation

I hope I found all cases. 

## ✅ Checks


internal/dependency/default_repo_adapter.go
```diff
@@ -13,41 +15,39 @@ type defaultRepoAdapter struct {
...
-	if err != nil {
-		return nil, err
+	packageRepo, repoErr := a.getRepoForPackage(name)
+	if repoErr != nil && packageRepo == nil {
+		return nil, repoErr
 	}
 	var idx repotypes.PackageIndex
 	if err := a.client.ForRepo(*packageRepo).FetchPackageIndex(name, &idx); err != nil {
-		return nil, err
+		return nil, multierr.Append(err, repoErr)
```
kosmoz, line 24: Nit: I think it's common practice to order the errors from oldest to newest
```suggestion
		return nil, multierr.Append(repoErr, err)
```

internal/dependency/default_repo_adapter.go
```diff
@@ -13,41 +15,39 @@ type defaultRepoAdapter struct {
...
-			return nil, fmt.Errorf("%v is available from %v repositories (currently unsupported)", name, len(repos))
-		}
+	repos, err := a.client.Meta().GetReposForPackage(name)
+	switch len(repos) {
+	case 0:
+		return nil, multierr.Append(fmt.Errorf("%v is not available in any repository", name), err)
+	case 1:
+		return &repos[0], err
+	default:
+		return nil, multierr.Append(fmt.Errorf("%v is available from %v repositories (currently unsupported)", name, len(repos)), err)
```
kosmoz, line 50: see previous comment
```suggestion
		return nil, multierr.Append(err,
			fmt.Errorf("%v is available from %v repositories (currently unsupported)", name, len(repos)),
		)
```

kosmoz (commented):
It works well! One thing I noticed is that there is an error toast on the list page, but not on the details page. I think it would be good to have at least some indication that not all is good on the details page as well. 

Code looks clean, but I see that there is a lot of `err != nil && mf == nil` and similar expressions. Would it make sense to create a custom error type that expresses the difference between a complete and a partial failure and shift the logic about what is a partial failure and what isn't to the metaclient? Just thinking here, let me hear your thoughts.

kosmoz:
> But wouldn't the caller then always have to first check for the error type and then do something like `partialErr.IsCompleteFailure()`? I think in some way, the caller will always have to do some logic about handling it.

Hmm, yes and no… the metaclient could use different error types for partial and complete failures. Thus, something like this might be possible:
```go
import repoerror "github.com/........."

if partial := (&repoerror.PartialError{}); errors.As(err, &partial) {
	// ...
} else {
	// assume non-partial
}
```

Alternatively, we could go the `apierrors` route with something like this:
```go
import repoerror "github.com/........."

if err := metaClient.GetRepoForPackage("..."); repoerror.IsPartial(err) {
	// ...
}
```

It is true that the caller still has to do the check but the underlying logic would be more centralized.

internal/repo/client/metaclient.go
```diff
@@ -101,6 +103,6 @@ func (d metaclient) GetReposForPackage(name string) ([]v1alpha1.PackageRepositor
 				}
 			}
 		}
-		return result, compositeErr
+		return result, repoerror.Partial(compositeErr)
```
kosmoz, line 106: I think we could go one step further and only return a partial error if at least one repo succeeded (and at least one failed).

internal/repo/error/repoerror.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"errors"
+)
+
+type PartialErr struct {
+	err error
+}
+
+func Partial(err error) *PartialErr {
+	return &PartialErr{err: err}
```
kosmoz, line 12: Not super important here, but for safety I think it would be good to check for nil here
```suggestion
	if err == nil {
		return nil
	}
	return &PartialErr{err: err}
```

internal/repo/error/repoerror.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+func (e *PartialErr) Error() string {
+	return e.err.Error()
+}
+
+func IsPartial(err error) bool {
+	var e *PartialErr
+	return err == nil || errors.As(err, &e)
```
kosmoz, line 21: Let me know if you disagree, but IMO an `IsSomething(err)` function should only ever return true if there is actually an error

kosmoz (changes requested):

internal/repo/error/repoerror.go
kosmoz: I think providing a suite of the following utils should be good:
- `IsPartial` --- err is of type PartialError
- `IsPartialOrNil` --- `err == nil || IsPartial(err)` 
- `IsTotal`/`IsComplete`/`IsNonPartial` --- `err != nil && !IsPartial(err)`

That should solve all required cases, right?

kosmoz (approved):

PR #1248, opened 2024-09-17
fix(ui): reopen strategy for portforwardings

### LINKED ISSUE #550: Open via UI fails when pod changes
**Describe the bug**
When during the same UI process a package has been opened, uninstalled, then installed again and opened again, the open fails. The tab is being opened, but it stays empty. There is also no error message on the UI, but one in the server console:

```
E0423 13:10:02.367041   75118 portforward.go:409] an error occurred forwarding 3000 -> 80: error forwarding port 80 to pod 1157c257cc625a8eaac376e63d2bb482f9da8fe66b9c891319d5e72e2896b8b0, uid : failed to find sandbox "1157c257cc625a8eaac376e63d2bb482f9da8fe66b9c891319d5e72e2896b8b0" in store: not found
```

There might also be other triggers which cause the forwarded-to pod to be deleted (e.g. config changes?). 

**To Reproduce**
1. Open a package via UI
2. Uninstall the package
3. Install the package again
4. Open the package via UI

**Expected behavior**
Open works. 

 - Type: Kubernetes
 - Version: 1.28
 - Provider: minikube

 - Package Operator version: v0.2.1
 - CLI version: v0.2.1

christophenne:
Closes #550 

## 📑 Description
Open to discuss if it should be done like that. An alternative could be to watch the forwarded-to pods and react to state changes – but that also would not cover all possible cases in which the portforwarding goes down, especially not the error cases. 

## ✅ Checks

### DIFF
--- a/internal/web/server.go
+++ b/internal/web/server.go
@@ -516,6 +516,8 @@ func (s *server) open(w http.ResponseWriter, r *http.Request) {
 
 func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrlpkg.Package) {
 	fwName := cache.NewObjectName(pkg.GetNamespace(), pkg.GetName()).String()
+	s.forwardersMutex.Lock()
+	defer s.forwardersMutex.Unlock()
 	if result, ok := s.forwarders[fwName]; ok {
 		result.WaitReady()
 		_ = cliutils.OpenInBrowser(result.Url)
@@ -526,9 +528,7 @@ func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrl
 	if err != nil {
 		s.sendToast(w, toast.WithErr(fmt.Errorf("failed to open %v: %w", pkg.GetName(), err)))
 	} else {
-		s.forwardersMutex.Lock()
 		s.forwarders[fwName] = result
-		s.forwardersMutex.Unlock()
 		result.WaitReady()
 		_ = cliutils.OpenInBrowser(result.Url)
 		w.WriteHeader(http.StatusAccepted)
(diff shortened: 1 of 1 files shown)


internal/web/server.go
```diff
@@ -525,10 +526,62 @@ func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrl
 	if err != nil {
 		s.sendToast(w, toast.WithErr(fmt.Errorf("failed to open %v: %w", pkg.GetName(), err)))
 	} else {
+		s.forwardersMutex.Lock()
```
kosmoz, line 529: Technically, I think this lock should be acquired before trying to read the forwarder the first time (line 519), to prevent Open being called multiple times. Will have to unlock in error case as well though

internal/web/server.go
```diff
@@ -525,10 +526,62 @@ func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrl
...
+			ctx = context.WithoutCancel(ctx)
+		resultLoop:
+			for {
+				select {
+				case <-s.stopCh:
+					break resultLoop
+				case fwErr := <-result.Completion:
+					// note: this does not happen in "realtime" (e.g. when the forwarded-to-pod is deleted), but only
+					// the next time a connection on that port is requested, e.g. when the user reloads the forwarded
+					// page or clicks open again – only then we will end up here.
```
kosmoz, line 546: Yes, this is to be expected. Kubectl exhibits the same behavior

internal/web/server.go
```diff
@@ -525,10 +526,62 @@ func (s *server) handleOpen(ctx context.Context, w http.ResponseWriter, pkg ctrl
...
+					if fwErr != nil {
+						fmt.Fprintf(os.Stderr, "forwarder %v completed with error: %v\n", fwName, fwErr)
+					} else {
+						fmt.Fprintf(os.Stderr, "forwarder %v completed without error\n", fwName)
+					}
+					// try to re-open if the package is still installed
+					if pkg.IsNamespaceScoped() {
+						var p v1alpha1.Package
+						if err := s.pkgClient.Packages(pkg.GetNamespace()).Get(ctx, pkg.GetName(), &p); err != nil || !p.DeletionTimestamp.IsZero() {
+							fmt.Fprintf(os.Stderr, "not reopening %v because of error/deletion: %v\n", fwName, err)
```
kosmoz, line 556: Minor but I think in case of a NotFound error, we can silently break the loop

kosmoz (approved):
> Open to discuss if it should be done like that. An alternative could be to watch the forwarded-to pods and react to state changes – but that also would not cover all possible cases in which the portforwarding goes down, especially not the error cases.

I think this approach is good. This is exactly how portforward is designed to work. In general the feature works as advertised, but because the reconnect only occurs after a failed connection attempt the browser still receives an error (though it works after a reload, of course). For now, this is definitely an improvement. 

I think a more sophisticated solution could involve a "proxy" listener that forwards data to the forwarder and waits in case it has to reconnect.

PR #1253, opened 2024-09-19
feat(ui): improve package detail page

### LINKED ISSUE #1167: Improve package detail page
As shortly discussed, we want to have all input data in all cases in one form. 
* The labels of the package detail header will become part of the form. The package detail header then only shows the status (will have to be tried out how that looks, if there is only one label – TBD).
* When advanced options are enabled or the package is not installed, repo and version are changeable, otherwise they are read-only.
* Warnings about advanced options have to be adjusted and should be next to the two fields that can be changed, if enabled.
* The new "header" part of the form (containing repo, version and auto updates) will be in the same form as the configuration below.
* Header and config part of the form should be visibly separated, but it should be clear that the save/install button on the bottom is meant to be for the entire page. 

For that, the template and the endpoints will have to be refactored a bit, and maybe the advanced option information needs to be stored somewhere other then local storage?

Also, be aware that this might (?) affect the live reload of the package detail page (there is some logic whether only header or whole page should be refetched). 

From a technical point of view, it would be interesting to explore some options for more client-side interactivity (see input config component madness, namespace check, etc.).

Ref #1091

christophenne:
Closes #1167 

Not a strict requirement, but maybe it makes sense to merge these before testing/merging this one:
* [x] #1254 
* [x] #1259 

## 📑 Description
* Repository and Version inputs are now shown always, but the selects are disabled when advanced mode is not enabled. 
  * When advanced mode is enabled, and a different manifest is chosen for an installed package (by changing repo and/or version), then the explicitly selected manifest is used to render the page.
  * If for an installed package, the manifest is changed, and the value definitions are not the same between the two manifests, there is a warning at the bottom, indicating that a value definition will get lost if the new manifest is being used. 
* Namespace and Name inputs are also shown always for namespaced packages, but disabled once it is installed. 
* These general inputs + the dynamic package config are visibly together in one form. 

Some refactoring has taken place too:
* advanced options are now stored in a cookie, because that way we have that info on the backend and can render accordingly. It's also way more straightforward to implement than with JS+localStorage and should have been done like this in the first place. 
* removed unused endpoints and moved the installation/configuration to the same path as the package detail page. 
* for installation/configuration, the form values of the package could have clashed with the hardcoded ones (repositoryName, etc), so I prefixed them with `values.`.
  * also changed some of the hardcoded names (`selectedVersion` -> `version`) to have the same names as in the package resource

## ✅ Checks

## ℹ Additional Information
A theme fix was made in https://github.com/glasskube/theme/pull/89

### DIFF
--- a/internal/web/cookie.go
+++ b/internal/web/cookie.go
@@ -26,7 +26,7 @@ func setAdvancedOptionsCookie(w http.ResponseWriter, advancedOptionsEnabled bool
 		MaxAge:   60 * 60 * 24 * 365,
 		Secure:   true,
 		HttpOnly: true,
-		SameSite: http.SameSiteLaxMode,
+		SameSite: http.SameSiteStrictMode,
 	}
 	http.SetCookie(w, &cookie)
 }
--- a/internal/web/templates/pages/package.html
+++ b/internal/web/templates/pages/package.html
@@ -166,7 +166,7 @@
               <div class="mt-1 mb-2">
                 <div class="form-check ">
                   <input class="form-check-input" type="checkbox" name="enableAutoUpdate" id="enable-auto-update" />
-                  <label class="form-check-label ms-1" for="enable-auto-update">Automatically Update Version</label>
+                  <label class="form-check-label ms-1" for="enable-auto-update">Automatically keep up-to-date</label>
                 </div>
               </div>
 


internal/web/templates/pages/package.html
```diff
@@ -127,73 +119,107 @@ <h2 class="text-reset">
...
+                    <div class="form-text">Namespace will be created if necessary</div>
+                  {{ end }}
+                </div>
+              {{ end }}
+
+
+              <div class="mt-1 mb-2">
+                <div class="form-check ">
+                  <input class="form-check-input" type="checkbox" name="enableAutoUpdate" id="enable-auto-update" />
+                  <label class="form-check-label ms-1" for="enable-auto-update">Automatically Update Version</label>
```
kosmoz, line 169:
```suggestion
                  <label class="form-check-label ms-1" for="enable-auto-update">Automatically keep up-to-date</label>
```

internal/web/cookie.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func setAdvancedOptionsCookie(w http.ResponseWriter, advancedOptionsEnabled bool) {
+	cookie := http.Cookie{
+		Name:     advancedOptionsKey,
+		Value:    strconv.FormatBool(advancedOptionsEnabled),
+		MaxAge:   60 * 60 * 24 * 365,
+		Secure:   true,
+		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
```
kosmoz, line 29: not that it really matters, but shouldn't SameSite=Strict work as well in this case?

kosmoz (commented):
Looks fantastic! I just have a small wording suggestion and a question (see below)

kosmoz (approved):

PR #1270, opened 2024-09-23
docs: use a mermaid diagram for the dependency management algorithm

linkdd:
## 📑 Description

Diagrams are often easier to read and grasp than text. The dependency management algorithm is such an example.

I translated the textual description to a Mermaid diagram with an intuitive color coding:

 - Green means no change
 - Yellow means successful change
 - Red means failure

Here is a screenshot of what it looks like:

glasskube-diagram

## ✅ Checks

## ℹ Additional Information

### DIFF
--- a/website/docs/05_design/dependency-management.md
+++ b/website/docs/05_design/dependency-management.md
@@ -25,15 +25,7 @@ Start --> A("Does P require a<br>version range of D?")
 
 A -->|No| B("Does D exist?")
 
-B -->|Yes| C("Are there other packages<br>dependent on D?")
-
-C -->|No| Fulfilled["P -> D is fulfilled"]
-
-C -->|Yes| E("Do X and Y require<br>no version range of D?")
-
-E ---->|Yes| Fulfilled
-
-E ---->|No| Fulfilled
+B -->|Yes| Fulfilled["P -> D is fulfilled"]
 
 B ----->|No| F["Install D pinned in latest(D)"]
 
--- a/website/docs/05_design/dependency-management.md
+++ b/website/docs/05_design/dependency-management.md
@@ -18,10 +18,10 @@ The following decision tree states how the Package Operator is handling dependen
 **Abbreviations:**
 
 - **P** Package that is going to be installed
-- **D** Dependency that a package require
-- **DV** Dependency version of an already installed packaged
-- **PDV** Desired Dependency version of package *P*
-- **XDV** / **YDV** Already existing packages *X* / *Y* that also have a dependency on *D*
+- **D** Dependency that package *P* requires
+- **DV** Installed version of package *D*
+- **PDV** Version constraint for package *D* as defined in the dependency relation of package *P*
+- **XDV** / **YDV** Version constraint for package *D* as defined in the dependency relation of already installed package *X* / *Y* that also have a dependency on *D*
 
 ```mermaid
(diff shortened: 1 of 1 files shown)


pmig:
Thanks for the PR - I checked it out locally and added a couple of improvements: https://github.com/glasskube/glasskube/commit/def31060a8efde7760ce459c0827ca5a77674d76

 - renamed the fiel to `.mdx` so IDEs can give you a better expirience
 - used top down formating for the graph
 - removed blank lines
 - add abbreviations
 
 Let me know what you think :-)

website/docs/05_design/dependency-management.md
```diff
@@ -12,38 +12,67 @@ The following decision tree states how the Package Operator is handling dependen
...
+
+B -->|Yes| C("Are there other packages<br>dependent on D?")
+
+C -->|No| Fulfilled["P -> D is fulfilled"]
+
+C -->|Yes| E("Do X and Y require<br>no version range of D?")
+
+E ---->|Yes| Fulfilled
+
+E ---->|No| Fulfilled
```
kosmoz, line 36: I just had a chat with @christophenne and we came to the conclusion that this branch is unnecessarily complex and should be simplified
```suggestion
B -->|Yes| Fulfilled["P -> D is fulfilled"]
```

linkdd:
I seem to be unable to fetch your commit to add it to the branch. Could you send it to me in mail format maybe? (using `git format-patch`)

website/docs/05_design/dependency-management.md
```diff
@@ -12,38 +12,67 @@ The following decision tree states how the Package Operator is handling dependen
...
+
+H -->|Yes| N("Is DV inside PDV?")
+
+N -------->|Yes| Fulfilled
+
+N ---->|No| P("Is DV < PDV?")
+
+P --->|Yes| Q["P -> D not fulfilled<br><b>Dependency Conflict</b><br>Might be resolvable if XDV, YDV, and PDV overlap"]
+
+P --->|No| R["P -> D not fulfilled<br><b>Dependency Conflict</b><br>Not resolvable because P does not support D in DV yet"]
```
kosmoz, line 64: I think this branch can be simplified by reordering the conditions like this:

```
Is DB inside pdv?
| yes --> fulfilled
| no --> Is DV < PDV?
| | yes --> Are there other existing packages dependent on D requiring a version range?
| | | yes --> conflict maybe resolvable
| | | no --> conflict resolvable
| | no --> conflict not resolvable 
```

Let me know what you think!

kosmoz (commented):
Thanks @linkdd this looks really valuable! The chart seems to be quite faithful to the original, but on inspection I think it can actually be simplified quite a bit. 

In addition to that, I think the chart should be preluded with a short explanation of the nomenclature used:
* single letters represent packages
* a letter followed by a "V" represents the version that package is installed in 
* two letters followed by a "V" represents the version constraint defined in the dependency relation between the packages represented by the first two letters

website/docs/05_design/dependency-management.md
linkdd: Change done, now it looks like this:

image

kosmoz (commented):
The chart looks really good now :slightly_smiling_face:

kosmoz (approved):

Issue #399, opened 2024-03-19
Show pruned packages at update and uninstall


christophenne:
**Is your feature request related to a problem? Please describe.**
During every update or uninstall its possible that required packages are going to be uninstalled as well. The user does not see which packages are going to be removed at all.

**Describe the solution you'd like**
UI and CLI should list the pruned packages to the user. 

On the UI we currently have the browser-built-in confirmation, this will have to be changed to the same bootstrap modal as in installation and update. 

**Describe alternatives you've considered**

**Additional context**

PR #1294, opened 2024-10-01
feat(ui): combine package update and configuration

christophenne:
## 📑 Description
The initial goal was to get rid of server-side state, like the `updateTransactions` being stored in the `server` struct. On that occasion, and because we previously introduced combined update+config in #1253 for "advanced users", I refactored the regular update process to be integrated in the package detail page as well. That means that the update modal has no use anymore and has been removed. 

Open to discuss visuals, "flow", wording, etc. 

We should also do the remaining part of #399 – showing pruned items on update, I put it into triage again. 

## ✅ Checks


kosmoz (commented):
Looks good, IMO. I would only change the icon to [`arrow-repeat`](https://icons.getbootstrap.com/icons/arrow-repeat/), because it seems to be a more accepted symbol for "update". (The current one, I would rather interpret as "upload")

Example (gnome software):
Screenshot from 2024-10-08 12-05-36

Example (windows; [source](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/simplified-windows-update-settings-for-end-users/ba-p/1497215)):
image

kosmoz (approved):

PR #1317, opened 2024-10-07
feat: show pruned packages on update

### LINKED ISSUE #399: Show pruned packages at update and uninstall

christophenne:
Closes #399
Depends on #1294

## ✅ Checks


internal/web/templates/pages/package.html
```diff
@@ -259,18 +259,30 @@ <h2 class="text-reset">
...
+                        <li>{{ $.Manifest.Name }} ({{ $.SelectedVersion }})</li>
+                      {{ end }}
+                      {{ range .ValidationResult.Requirements }}
+                        <li>{{ .Name }} ({{ .Version }})</li>
+                      {{ end }}
+                    </ul>
+                  </div>
+                {{ end }}
+                {{ if and .ValidationResult.Pruned }}
+                  <div class="alert alert-danger m-0 mb-2" role="alert">
```
kosmoz, line 277: we normally use "danger" for errors and things that are not possible
```suggestion
                  <div class="alert alert-warning m-0 mb-2" role="alert">
```

kosmoz (changes requested):
It works! Just 2 small change requests

PR #1331, opened 2024-10-10
feat(package-operator): restart dependent workloads when package spec changes

### LINKED ISSUE #1260: Restart deployments when package spec changes
For example, when we patch environment variables most likely the containers need to be restarted in order to read the new config values. There might also be other occasions?

TBD.

christophenne:
Closes #1260 

## 📑 Description
For all owned deployments and stateful sets, the operator annotates their template with an annotation that contains the hash of the package spec. Anytime the package spec and therefore its hash changes, the annotation will be updated, causing the deployment/stateful set to do a restart. 

## ✅ Checks


internal/controller/ctrlpkg/util.go
```diff
@@ -1,7 +1,29 @@
...
+	"fmt"
+)
+
 func IsSameResource(a, b Package) bool {
 	return a.GetName() == b.GetName() &&
 		a.GroupVersionKind() == b.GroupVersionKind() &&
 		a.GetNamespace() == b.GetNamespace()
 }
+
+func SpecHash(pkg Package) (string, error) {
```
kosmoz, line 16: I think it would be fair to move this function to the api package, so it can be called as `pkg.GetSpec().Hashed()` or something like that

internal/controller/ctrlpkg/util.go
```diff
@@ -1,7 +1,29 @@
...
+	if specBytes, err := json.Marshal(pkg.GetSpec()); err != nil {
+		return "", fmt.Errorf("failed to marshal package spec: %w", err)
+	} else {
+		var currentSpecHash string
+		h := sha256.New()
+		if _, err := h.Write(specBytes); err != nil {
```
kosmoz, line 22: A slightly more efficient way to do this would be a streaming JSON encoder like `json.NewEncoder(h).Marshal(pkg.GetSpec())`. 

It probably wouldn't matter here performance wise, but I think it's more concise too, so I would personally prefer that way :slightly_smiling_face:

api/v1alpha1/well_known_annotations.go
```diff
@@ -3,4 +3,5 @@ package v1alpha1
 const (
 	AnnotationAutoUpdate     = "packages.glasskube.dev/auto-update"
 	AnnotationInstalledAsDep = "packages.glasskube.dev/installed-as-dependency"
+	AnnotationSpecHash       = "packages.glasskube.dev/spec-hash"
```
kosmoz, line 6: I think something like `package-spec-hashed` or `package-spec-sha256` would be a little more meaningful to someone unfamiliar with the internal mechanics

kosmoz (changes requested):
Ohh this is very clever, it means we don't need to put the hashed spec on the package resource itself. I like that a lot!

kosmoz (approved):

PR #1360, opened 2024-10-24
docs(website): added initial glossary layout and first term

jakepage91:
Opening the initial Glossary 
Screenshot 2024-10-25 at 15 12 28

## ✅ Checks


jakepage91:
operator
kubernetes-cluster-architecture

website/docs/08_glossary/01_devops.mdx
```diff
@@ -0,0 +1,44 @@
+---
```
pmig, line 1: All pages glossary entries should not be in the docs/ folder but rather in the glossary folder, so the url of a glossary entry is: not `http://localhost:3000/docs/glossary/devops/` but `http://localhost:3000/glossary/devops/`

website/docs/08_glossary/01_devops.mdx
```diff
@@ -0,0 +1,44 @@
...
+title: DevOps
```
pmig, line 2: all glossary pieces should also have a description (image might be nice, but can be added as step two)

website/docusaurus.config.ts
```diff
@@ -211,6 +223,10 @@ const config: Config = {
               label: 'Getting started',
               to: '/docs/getting-started/install',
             },
+            {
+              label: 'Glossary',
```
pmig, line 227: I wouldn't put the Glossary below docs, but rather in the top navigation bar below resources

website/docusaurus.config.ts
```diff
@@ -136,6 +136,18 @@ const config: Config = {
       },
     ],
   ],
+  customFields: {
+    glossaryItems: [
```
pmig, line 140: You could copy how guides a set up in that way a "category" page should be automatically generated.

pmig (changes requested):
Overall a good first version und Content of the glossary. It would have also been possible to copy the structure the glossary the same way guides are implemented in the website, but fixing the URL so that glossary pieces are not under docs, should work also, so the url will not change if we decide the refactor it later :-)

website/docs/08_glossary/01_devops.mdx
jakepage91: This is done

website/docusaurus.config.ts
jakepage91: Got it! Now it's under the `Resources` tab

website/docs/08_glossary/01_devops.mdx
jakepage91: I've added a description but will leave the image to another step since I'm not sure the best way to visualize it.

jakepage91:
I think it should be good to go now @pmig

pmig (approved):

PR #1383, opened 2024-11-05
fix(package-operator): also set `ownedPackages` when dependencies unresolved

### LINKED ISSUE #1261: Dependencies/Components not cleaned up when parent has status failed
**Describe the bug**
When uninstalling a failed namespaced package, it seems like some dependencies and/or components are not deleted, only the parent/root package is deleted.

**To Reproduce**
1. Install tracecat
2. configure it invalidly to have it in status "Failed"
3. uninstall it

The database and temporal (and therefore its components) are still in the cluster. 

**Expected behavior**
They should be cleaned up.

**Screenshots**

 - Type: [e.g. Kubernetes, OpenShift]
 - Version: [e.g. 1.29]
 - Provider: [e.g. AWS, minikube]

 - Package Operator version: latest dev
 - CLI version: latest dev

### LINKED ISSUE #1316: `ownedPackages` should be set earlier
**Describe the bug**
When installing a package with a dependency, the parent package usually goes to status "Pending" until the next reconciliation. During that time, the `ownedPackages` does not contain the required package yet, so if one deletes the package during its in Pending status, the required package will not get purged (it has the `installed-as-depdency` annotation though). 

Haven't debugged it yet, but could be similar to #1261?

Fix should probably be to set the `ownedPackages` already at the first reconciliation.

christophenne:
Closes #1316
Closes #1261

## ✅ Checks

### DIFF
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -482,7 +482,7 @@ func (r *PackageReconcilationContext) ensureDependencies(ctx context.Context) bo
 		}
 	}
 
-	ownerutils.Add(&r.currentOwnedPackages, ownedPackages...)
+	r.setShouldUpdate(ownerutils.Add(&r.currentOwnedPackages, ownedPackages...))
 
 	if len(failed) > 0 {
 		message := fmt.Sprintf("required package(s) not installed: %v", strings.Join(failed, ","))
--- a/internal/controller/common.go
+++ b/internal/controller/common.go
@@ -482,6 +482,8 @@ func (r *PackageReconcilationContext) ensureDependencies(ctx context.Context) bo
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controller/common.go
```diff
@@ -482,6 +482,8 @@ func (r *PackageReconcilationContext) ensureDependencies(ctx context.Context) bo
 		}
 	}
 
+	ownerutils.Add(&r.currentOwnedPackages, ownedPackages...)
```
kosmoz, line 485: I think we should also call `r.setShouldUpdate` here 

```suggestion
	r.setShouldUpdate(ownerutils.Add(&r.currentOwnedPackages, ownedPackages...))
```

kosmoz (approved):

PR #1396, opened 2024-11-12
refactor(ui): switch to golang router and restructure

christophenne:
## 📑 Description
The main goal was to refactor the routing such that we can let go of the `mux` dependency and use the [new features](https://go.dev/blog/routing-enhancements) builtin http router of Go. We didn't really use much of `mux` features anyway, except for reading the named path parameters, so using a big router library for that was a bit of an overkill. 

On that occasion, improvements in the template handling have been made, which were long overdue. Following problems were present so far:
* The startup code involved cloning a base template for every single template of the application. That was a decision taken at the very beginning – but it always felt a bit weird and wrong. In the refactored version, there is just one instance of a `template.Template`, that now contains all the pages and components. This also meant that for the pages, the template name changed from `"content"` (which would have been ambiguous), to `"pages/<pageName>"`. That way it is possible to execute a template by its name. For pages, in order to avoid the layout duplication for each page, we simply always render `"base.html"` and pass the given template name into it. 
* The `server.go` file contained a large portion of the business logic, and the file simply outgrew over the time. We put some of the things in separate files or even modules, but it was a bit of a mess.
* The endpoint handlers all were functions of the `server` struct, which has become sort of a "god object" which held all state. Now we have a `controllers` module, where the features are separated into files. The handler functions do not access the `server` object anymore, but instead use any needed dependencies (like the cached package client) via the `context.Context`. Other dependencies like the dependency manager or value resolver are now more being created on the fly (because they don't actually hold state). 
* The endpoint handlers had direct access to the templates, and they mostly put a `map[string]any` into them. Now the template rendering aspect has been abstracted with the `responder` package, which can still take an untyped map too, but in most cases we call it with a respective view struct. 

Further things: 
* the way to display a package manifest url in a template has changed from a template func to simply being passed from outside via the view object. Otherwise the template func has a "weird" dependency on the `repoClientset`.
* the check whether telemetry is enabled was previously done un-cached, and therefore when doing a lot of activity on the UI, could run into a client-side rate limiting. The already existing `namespaceLister` can now be passed to the telemetry client instead. 

## ✅ Checks

### DIFF
--- a/internal/clicontext/context.go
+++ b/internal/clicontext/context.go
@@ -6,23 +6,20 @@ import (
 	repoclient "github.com/glasskube/glasskube/internal/repo/client"
 	"github.com/glasskube/glasskube/pkg/client"
 	"k8s.io/client-go/kubernetes"
-	appsv1 "k8s.io/client-go/listers/apps/v1"
-	v1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd/api"
 )
 
 type (
-	contextKey int
+	ContextKey int
 )
 
 const (
-	pkgClientContextKey contextKey = iota
+	pkgClientContextKey ContextKey = iota
 	k8sClientContextKey
 	configContextKey
 	rawConfigContextKey
 	repoClientsetContextKey
-	coreListersContextKey
 )
 
 func SetupContext(ctx context.Context, config *rest.Config, rawConfig *api.Config) (context.Context, error) {
@@ -51,10 +48,6 @@ func SetupContextWithClient(
 
 func ContextWithRepositoryClientset(parent context.Context, clientset repoclient.RepoClientset) context.Context {
 	return context.WithValue(parent, repoClientsetContextKey, clientset)
-}
-
-func ContextWithCoreListers(parent context.Context, coreListers *CoreListers) context.Context {
-	return context.WithValue(parent, coreListersContextKey, coreListers)
 }
 
 func PackageClientFromContext(ctx context.Context) client.PackageV1Alpha1Client {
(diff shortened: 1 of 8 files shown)


internal/web/controllers/settings.go
```diff
@@ -0,0 +1,130 @@
...
+			responder.SendToast(w, toast.WithErr(fmt.Errorf("use a valid URL for the package repository (got %v)", err)))
+			return
+		}
+		repo.Spec.Url = repoUrl
+	}
+
+	repo.Spec.Auth = nil
+
+	if checkDefault == "on" {
+		// TODO "cliutils" in the server??
```
kosmoz, line 102: The "cli" part in `cliutils` means that it should not be used by the package operator.
Since the server is part of the "cli" (aka client application) this is OK. 

But I agree that the naming is kind of ambiguous here, since we have two distinctions:
1. CLI <=> operator
2. CLI <=> server

internal/web/controllers/packages.go
```diff
@@ -0,0 +1,127 @@
...
+		if pkg.ClusterPackage != nil {
+			installedClpkgs = append(installedClpkgs, pkg.ClusterPackage)
+		}
+		clpkgUpdateAvailable[pkg.Name] = isUpdateAvailableForPkg(ctx, pkg.ClusterPackage)
+	}
+
+	overallUpdatesAvailable := false
+	if len(installedClpkgs) > 0 {
+		overallUpdatesAvailable = isUpdateAvailable(ctx, installedClpkgs)
+	}
```
kosmoz, line 48: shouldn't `isUpdateAvailable` called with an empty slice return false anyways?

internal/clicontext/context.go
```diff
@@ -99,3 +106,21 @@ func RepoClientsetFromContext(ctx context.Context) repoclient.RepoClientset {
 	}
 	return nil
 }
+
+// TODO too web specific and should maybe be an web-extension ?
```
kosmoz, line 110: I think I agree. Maybe we can put this into a new `internal/web/context` package?

kosmoz (commented):
I have some comments, but I think overall this is a really good improvement!

kosmoz (approved):

PR #1419, opened 2024-11-25
docs(website): fixed broken glossary links

jakepage91:
I realised that the three newly added glossary pieces had broken links.

## ✅ Checks


pmig (changes requested):
I would prefer having the full slug in the url

Issue #405, opened 2024-03-20
Build numbers not considered in dependency ranges


christophenne:
**Describe the bug**
When defining a dependency with a version range which includes a build number, e.g. ">= 1.13.3+2", the build number is ignored and version 1.13.3+1 would also be considered as compatible. 

**To Reproduce**
1. use local package repo, where kubernetes-dashboard depends on cert-manager > 1.13.3+1
2. install cert-manager in version 1.13.3+1
3. install kubernetes-dashboard: cert-manager will not be updated, although it would be required when installing kubernetes-dashboard

**Expected behavior**
Up for discussion: When checking a version against a version constraint, the build number should be considered too. Alternatively, we can also disallow using build numbers in the dependency version ranges, which would make things a lot easier. 

**Glasskube Info (please complete the following information):**
 - Package Operator version: dev
 - CLI version: dev

**Additional context**
Similar to #397

pmig:
@christophenne can this ticket be closed after https://github.com/glasskube/glasskube/pull/426 got merged?

christophenne:
No this is not the same. #426 fixes #397, but afterwards we realized we have the same problem when declaring versions in dependencies. It's up for discussion how we want to deal with it (see expected behavior).

christophenne:
Decision: won't fix but document the constraint

PR #1561, opened 2026-05-08
fix: reject build metadata in version constraints

immanuwell:
Refs #405, #1276

## Description
Dependency constraints with build metadata are documented as unsupported, but the graph parser still accepted them.

Before this patch, `<= 1.1.0+1` behaved like `<= 1.1.0`, so `1.1.0+2` could still pass. small footgun, kinda sneaky.

## Reproduce
Before the fix, the graph test expected `1.1.0+2` to pass under a `<= 1.1.0+1` constraint:

```sh
go test ./internal/dependency/graph
```

Now dependency and component constraints go through the local semver helper and reject build metadata up front.

## Checks
`go test ./internal/semver ./internal/dependency/graph`

`go test ./...`

No docs update needed, the docs already say build metadata in dependency ranges is unsupported.

### DIFF
--- a/internal/dependency/graph/graph.go
+++ b/internal/dependency/graph/graph.go
@@ -276,7 +276,7 @@ func (g *DependencyGraph) add(
 
 		var constraint *semver.Constraints
 		if len(dep.Version) > 0 {
-			if c, err := semver.NewConstraint(dep.Version); err != nil {
+			if c, err := isemver.NewConstraint(dep.Version); err != nil {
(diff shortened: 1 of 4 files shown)


pmig:
Thanks for your contribution, we are not planning on releasing new versions for the package manager, please consider forking it if you are actively using it.

immanuwell:
@pmig got it, so I'm going to close this PR

